PR #58, opened 2015-11-26
added Dockerfile

cahna:
I created this to deploy in my home setup and thought I'd see if you'd be interested in accepting it. 

If so, I'd suggest linking the repository to docker hub as an automated build so the docker image will always be up-to-date :smiley: 

Here's an example setup:
```yaml
# docker-compose.yml
---
motioneye_data:
  image: cheine/motioneye
  entrypoint: /bin/true
  volumes:
    - /docker/motioneye/data:/var/lib/motioneye
    - /docker/motioneye/pids:/var/run/motioneye
    - /docker/motioneye/config:/etc/motioneye

motioneye:
  image: cheine/motioneye
  restart: always
  volumes_from:
    - motioneye_data
  ports:
    - 8765:8765
```

P.S. Great work! I'm so happy to have found this. Goodbye forever, zoneminder.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,52 @@
+
+FROM ubuntu:15.04
+MAINTAINER Conor Heine <conor.heine@gmail.com>
+
+ENV DEBIAN_FRONTEND noninteractive
+ENV export LANGUAGE=en_US.UTF-8
+ENV export LC_ALL=en_US.UTF-8
+ENV export LANG=en_US.UTF-8
+ENV export LC_TYPE=en_US.UTF-8
+
+RUN apt-get update
+
+# Core
+RUN apt-get --yes install \
+        git \
+        motion \
+        ffmpeg \
+        v4l-utils \
+        python-pip \
+        libssl-dev \
+        libjpeg-dev \
+        libcurl4-openssl-dev
+
+ # Python
+RUN apt-get --yes install \
+        python2.7 \
+        python-setuptools \
+        python-dev \
+        python-pip
+
+# Pip
+RUN pip install tornado jinja2 pillow pycurl
+
+# Fetch motioneye src
+RUN cd /tmp && git clone https://github.com/ccrisan/motioneye.git && \
+    cd /tmp/motioneye && python setup.py install && mkdir /etc/motioneye && \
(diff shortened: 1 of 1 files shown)


ccrisan:
Well, I don't see why not. Thanks for your contribution. However, I'll move the `Dockerfile` to the `extra` folder (I hope that won't break anything - I don't have experience with Docker at all).

cahna:
Happy to contribute. The `Dockerfile` doesn't depend on any build context, so the `extra` folder will work.

BTW, I'm a python developer with a lot of free time in the next few weeks. I'd be happy to spend some of that time on motioneye if you have any features/bugs outstanding.

ccrisan:
@cahna It's great to hear that you speak python, you have some free time and you're willing to help. These three conditions have never been met before by anyone :)

Actually users have been asking me a lot to add audio support to the whole solution. Do you think you would be up to this task? Normally there are two subtasks when it comes to audio: (1) real-time transmitting of audio stream along with the video one (no recording, no touching of motion daemon) and (2) adding an audio track to the recorded movie files (involves a lot of work and dealing with motion C source code). What do you think?

cahna:
Sounds like a proper challenge, @ccrisan. Having audio support would be nice. Let me do some research on motion and read the motioneye code to get a better idea of what needs to be done. If you don't have any imminent deadline requirements, I'd be happy to give it a shot. 

Want to take further discussion to [gitter](https://gitter.im), irc, or email? I know I'll have questions.

ccrisan:
(Un)fortunately there aren't any urgent issues; if there had been any, I would have gotten them fixed already :)

Regarding further discussions, sure, we could use any of them. I never used gitter before, but it seems quite nice. Let's give it a try.

rickerd:
@cahna can you point out how to run the docker. I want to contribute to implement better use of the UI.
I don't speak Python, sadly, but i do wana contribute! :)

jeff-h:
I've added some notes to the Wiki on running motionEye in Docker.

https://github.com/ccrisan/motioneye/wiki/Install-in-Docker

Hard2findaNameHere:
Hi Calin, thank you for your previous help with the action buttons, my little project is working good now.   I noticed that your road map includes the possibility of audio.  Has there been any progress on this?

PR #65, opened 2015-12-07
Added more resolution options

algestam:
Added resolution options matching the resolutions available on the QD900 IP Camera.


ccrisan:
1080 % 16 != 0, which is a requirement for motion. The 1600x1200 one is alright.

algestam:
Ok... I tested it myself and it worked fine. The reason that it works for
me is probably that I have been using the Mr-Dave fork of motion (
https://github.com/Mr-Dave/motion) in order to get rtsp streams to work.
Apparently, his fork also contains a fix for other resolutions (
https://github.com/Mr-Dave/motion/issues/28).


2015-12-08 15:34 GMT+01:00 Calin Crisan <notifications@github.com>:

ccrisan:
Interesting, I did not know about that change. Thanks for letting me know.

algestam:
Would you be willing to accept a change to motioneye where the new resolutions are added only if the motion binary is supporting it?

I'm thinking about checking the version of motion and only if it is from the MrDave or sackmotion, the new resolutions would be added. If you have any ideas on how to do this in a better way, I am all ears :)

ccrisan:
Yes, that's the best way to implement it. I already have a few check functions that tell things about motion capabilities; it wouldn't hurt adding another one about the resolution.

cvpeck:
Is this one going to be merged? I raised this as a feature request before I saw this pull request was waiting. I currently hack the config files by hand to support the 1920x1080 output by the Foscam cameras. Fork of motion with H.264 support used as well.

speeskeek:
hi there. I also manually edit the config (thread-1 thread-2 ...) to have the resolution of my choice.
Unfortunately, every time I edit something in the webUI the resolution is overwritten.
I'm using D1 resolution, or something similar my cheap china cams produce (704x578 or 704x482).

It's working great like this, so maybe add a manual resolution entry, but with a warning this might not work with certain motion builds?

algestam:
Closing this pull request since modulo 8 resolutions now are valid in motioneye (with a version of motion supporting it). Thanks for the fix :)

PR #95, opened 2016-01-06
Update mediafiles.py

mvanbreeden:
Remove .AppleDouble directories from media file paths to prevent errors in thumbnail and previews


ccrisan:
Fixed with https://github.com/ccrisan/motioneye/commit/8c142b1b091aad180a61893630948913c814c3dd.

PR #137, opened 2016-02-26
Be able to save an image by right clicking (or long tap).

nmrugg:
Right now, if you right click to try to locally save an image from a camera, it saves the progress loader instead. This change lets you click (or tap) and save a frame.


ccrisan:
Thanks. This is indeed something that has been bothering me as well. I must admit I did not know about the `pointer-events` CSS property!

PR #186, opened 2016-04-20
add checkboxes to individually control upload of pictures vs videos

bovine:
This adds the ability to enable file uploading of only certain types of files (photos or movies). The default configuration is to upload both types, which is what motioneye currently does.

This functionality might be especially useful for people that only want to upload one type, but still capture both types locally for reconstruction purposes.

upload-types


ccrisan:
Reviewed and merged. Thanks!

PR #218, opened 2016-06-06
Movie playback

dermotduffy:
Hi,

**TL;DR:** Add in browser playback support.

At least Mr-Dave's fork of motion trivially supports encoding in mp4 (via 'ffmpeg_video_codec mp4' in your motion configuration). This pull request enables playback of the movies in browser, if the browser supports them, via HTML5 <video> tags. This necessitated adding 'Content-Range' support to the MovieHandler, which browsers use to seek the content. 

Testing:
- OSX: Safari, Firefox, Chrome.
- Android.
- IE: Windows.
- Also basic testing of 'Remote MotionEye' cameras (which I do not use) by connecting two motioneyes together. This quite likely does not include all use cases of the remote functionality (that I do not know about).

Longer term it might be better to use StaticFileHandler directly to serve these (and wrap remote content somehow). It would also be good to add mp4 encoding to the UI as an option if the version of motion supports it -- but I figured I'd start here to see how you feel about this.

Thanks,
Dermot.


ccrisan:
This is some really nice work! I'll take a look over it as soon as I get some spare time.

motioneye/static/js/main.js
```diff
@@ -2919,15 +2919,23 @@ function runPictureDialog(entries, pos, mediaType) {
     
     var img = $('<img class="picture-dialog-content">');
     content.append(img);
-    
+
+    var video_container = $('<video class="picture-dialog-content" controls="true">');
+    var video_source = $('<source type="video/mp4">')
```
PaulAnnekov: Remove this and use `src` directly on `<video>` or pass mime type from backend and insert it here.

motioneye/static/js/main.js
```diff
@@ -2941,29 +2949,55 @@ function runPictureDialog(entries, pos, mediaType) {
         
         prevArrow.css('display', 'none');
         nextArrow.css('display', 'none');
+
+        /* Construct a likely mime-type with 'video/' and the file extension, then see if the 
+           browser can play it */       
+        var playable = video_container.get(0).canPlayType('video/' + entry.path.split('.').pop()) != ''
```
PaulAnnekov: Just pass mime type from backend and use it here.

motioneye/handlers.py
```diff
@@ -1338,28 +1340,78 @@ def download(self, camera_id, filename):
...
+        range_header = self.request.headers.get("Range")
+        if range_header:
+            request_range = httputil._parse_request_range(range_header)
+
         if utils.local_motion_camera(camera_config):
-            content = mediafiles.get_media_content(camera_config, filename, 'movie')
-            
+            full_path = os.path.join(camera_config.get('target_dir'), filename)
+            size = os.stat(full_path).st_size
+            if request_range:
```
PaulAnnekov: Manual `Range` manipulation is not a good idea. @ccrisan maybe your python server has such feature?

motioneye/static/js/main.js
```diff
@@ -2941,29 +2949,55 @@ function runPictureDialog(entries, pos, mediaType) {
...
         progressImg.css('left', (img.parent().width() - progressImg.width()) / 2);
         progressImg.css('top', (img.parent().height() - progressImg.height()) / 2);
         
         img.attr('src', addAuthParams('GET', basePath + mediaType + '/' + entry.cameraId + '/preview' + entry.path));
+
+        if (playable) {
+            playButton.on('click', function() {
+                video_source.attr('src', addAuthParams('GET', basePath + mediaType + '/' + entry.cameraId + '/download' + entry.path));
+                video_container.show();
+                video_container.get(0).load();  /* Must call load() after changing <video> source */
```
PaulAnnekov: Errors must be handled. VideoJS will handle them instead of you showing nice message. Or you can just add `error` listener add display error message to user instead of silent, broken player.

PaulAnnekov:
@ccrisan have you found some time to review this?

motioneye/handlers.py
ccrisan: Unfortunately I have only read about this http range functionality "in books" and never actually had to deal with it. I still need to allocate some serious amount of time to take a look over all this code.

ccrisan:
Nope, unfortunately I haven't managed to take a serious look over your work yet, but given that more and more users request this feature, I really need to take this into consideration asap. It's next on the list after motion detect mask support, upon which I have already started working in a separate branch.

Anyways, thanks a lot for your contribution and I promise I'll do my best to integrate this into mainline, as it's some valuable piece of work.

motioneye/handlers.py
PaulAnnekov: I think we can remove it now. The video will be unseekable until fully downloaded, but we will release it faster and can handle this in next PR. It still be much usable then now.

dermotduffy:
Sure. Thanks for the comments, appreciated. Let me review them, make some fixes and consider splitting the PR as you suggest.

ccrisan:
@dermotduffy, @PaulAnnekov my apologies for mistaking the real author of the pull request!

dermotduffy:
All comments from @PaulAnnekov have been addressed (thanks Paul). The removal of Range support makes the PR significantly simpler, at a cost of losing movie seeking. 

Hopefully this helps get this feature in sooner. If this PR gets pulled in, I'll submit a followup PR to add seeking back in -- that is a much smaller feature difference than being able to play movies vs not being able to play movies, so it can afford to wait longer (I think this is the point Paul is making).

motioneye/static/js/main.js
```diff
@@ -2919,15 +2919,24 @@ function runPictureDialog(entries, pos, mediaType) {
     
     var img = $('<img class="picture-dialog-content">');
     content.append(img);
-    
+
+    var video_container = $('<video class="picture-dialog-content" controls="true" type="video/mp4">');
```
PaulAnnekov: `type` is redundant. It makes nothing. Only `<source>` has this param.

motioneye/static/js/main.js
```diff
@@ -2919,15 +2919,24 @@ function runPictureDialog(entries, pos, mediaType) {
...
+    video_container.on('error', function(err) {
+      showErrorMessage('Error: Unable to play movie.')
```
PaulAnnekov: `err` has error code, that will be much useful to understand exact problem. Here is a list of errors and messages: https://gist.github.com/wilmoore/3252894. We can handle it in next PR.

motioneye/static/js/main.js
```diff
@@ -2941,29 +2950,55 @@ function runPictureDialog(entries, pos, mediaType) {
         
         prevArrow.css('display', 'none');
         nextArrow.css('display', 'none');
+
+        /* Construct a likely mime-type with 'video/' and the file extension, then see if the
```
PaulAnnekov: This message is not relevant anymore :)

PaulAnnekov:
@dermotduffy great! I have left some small comments.

peppelinux:
This is a very important  feature, I'm still waiting for it.
Could I be helpfull for merge ?

Another important feature is the possibility to select more then one picture/movie and then delete them in one shot.

dermotduffy:
[Merged master back in]

@ccrisan your the mask support now nicely submitted, do you think you'd have time to look at this?

With the improvements @PaulAnnekov suggested, including the much simplified http handling (i.e. with the Range support removed for now) I think this should be a pretty straightforward review and feature to test -- feels like a good return on review investment for a feature that seems to be desirable for multiple people.

ccrisan:
Next in line is API support and HTTP authentication (as an alternative to the existing signature-based one). After that I'll take care of the HTML5 video playback feature.

bah2830:
@dermotduffy Thank you for this. I have merged your branch into my local repo for the time being until this can be merged into master. Everything is working well.

It does have an interface issue when a video is clicked that isn't completed recording. The div resizes to be a bit taller. Then if the previous button is pressed and it loads an older video the size does not reset causing the aspect ratio to be off. This is likely an artifact of the frontend framework and wont be solved with this PR but thought I should let you know.

ChriZathens:
Are there any plans to implement this in the short future?

bobhruska:
I've been using this with much success but have recently come across a problem.  For some reason the way the video is presented to the browser does not allow it to work on my iphone, either running chrome or safari.  The error received is "Media format unsupported or otherwise unavailable/unsuitable for playing"

It is not a code/container issue because I can directly link to a copy of the video on my server and it will play on the phone.

Any ideas?

dermotduffy:
Wild guess is because the Range support that the browser may expect is unavailable. It was removed to make this PR simpler thus notionally increasing the chances of it getting submitted to this repo. You could try restoring the code stripped out here (https://github.com/ccrisan/motioneye/pull/218/commits/b3ea5371d0d0817e4ec9476a7036e3630cf900be) to see if that works better for you.

bobhruska:
I did a quick manual merge of the old handlers.py and remote.py and it now plays on my iphone!
Thanks for the quick reponse.

Any idea what needs to be done to get this playback feature put into the master?

dermotduffy:
That's a question for @ccrisan -- on where this currently lives on his priority list.

As the PR is more than a year old, I'm suspecting it may not happen: perhaps it's just not important enough to the author, there isn't enough time, or there are other incompatible plans that I don't see, etc. I  do think it's a pretty useful feature and certainly the most obvious gap I saw when I started using MotionEye.

If this does come up for review, I'm happy to (at author's discretion) re-add the Range support. Whilst it makes the PR a little more complex to review, the end result is a much better user experience (and apparently a working iPhone).

ccrisan:
I haven't had time to look over this yet. I'll try to find some spare time to review the PR asap.

rozza-m:
For me this is the one obvious omission / opportunity with MotionEye.

@dermotduffy - as @ccrisan has been struggling over the last two years to find time to review, maybe you'd be prepared to re-review your code, fork motioneye and release binaries to include this improvement?

sv-kov:
@dermotduffy  I also join the request, you can correct your code and put the changed files.

vganyn:
@dermotduffy join too

WayneKeenan:
=2

kristian:
Also my Kudos would go for resolving the merge conflicts & merging this one please!

noloco:
This is an essential feature of any surveillance software.  +1

themouseNZ:
+1 Please would make this great software even better

flargargoyl:
Tested this changes in my Motioneye installation. Is it just me having some issues with playback of .mp4 or someome had it too? I`m using h.256 & ffmpeg obviously ,yet a waiting time before playback is like more than 3-5 minutes for a single movie

Nukem36:
+1 from my side for this request.

ShanDestromp:
For those interested, this method while not recently updated does still work, although you'll need to manually merge the changes as the files have been changed.

Additionally I had to make a few minor changes in main.js to get playback to actually work as I kept getting errors in the console.  I have virtually zero knowledge of JS so it was simply a matter of Googling the error and making tweaks until it worked.

Attached below is a ready-to-go version of main.js based off of available main source (as of a few days ago at least).  All the other files can be manually merged fairly easily although line numbers might have changed.

Similar to @flargargoyl however I'm experiencing a delay before playback is available (though with h.264 files).  Regardless of what `preload` is set to it seems as though the entire file is being forced to buffer, opposed to starting playback once a minimum amount is retrieved.

[main.js.txt](https://github.com/ccrisan/motioneye/files/1579385/main.js.txt)

darrenparkinson:
Keen to get this implemented.  It's a real show stopper not being able to play recorded videos on a mobile device.  Any progress on it?  Many thanks.

rasssta:
+1 on this feature as well. It's pretty much the only thing I'm missing in this application. Great work!

motioneye/handlers.py
```diff
@@ -1441,13 +1442,12 @@ def download(self, camera_id, filename):
         camera_config = config.get_camera(camera_id)
         if utils.is_local_motion_camera(camera_config):
             content = mediafiles.get_media_content(camera_config, filename, 'movie')
-            
+
```
jasaw: There's a lot of space removal changes like this one. They should not be part of this pull request.

jasaw:
The changes look quite straight forward. Has anyone done any browser compatibility tests? Which versions of which browsers have been tested?

themouseNZ:
Can it not be merged and we can let you know?

dermotduffy:
It's 1.5 years since I first prepared this PR. I'm not sure what to do with this? It does not appear to be a priority for the project owner to pursue even after the changes were largely refactored to reduce review burden. @ccrisan , any thoughts on how you'd like me to proceed? I'm open to anything, including just closing this PR -- but I don't plan to invest the time to keep this change clean/mergeable indefinitely if it doesn't stand a reasonable change of actually being merged in.

My only alternative suggestion is that perhaps someone else who wants this feature and who has the time maintains a fork with proper movie playback. The changes as they stand, and the earlier removed Range support, provide a decent starting point.

ccrisan:
@dermotduffy this is by far the most requested feature. I must admit I did not have too much time to spend on motionEye lately and I wanted to take care of the simpler problems first. Moreover, there were at least 3 other users (apart from you) that have offered to implement this feature in the past 12 months or so. They didn't come through though - they might still be working on it :).

I must apologize for neglecting this PR, despite the highly demanded feature it implements. I was reluctant to the way it would handle the many video codes/formats motion currently supports and also the browser support.

Anyways, let's see what we can do to get this merged. **I promise to make this a priority**, to review what needs to be reviewed in due time and to get it merged asap. So here are my questions:

1. Given the current PR, are there any compatibility issues with the recent motion versions (and their supported video codecs/formats)?
2. What is the browser compatibility state as of right now (I know next to nothing about HTML5 video)?
3. Do we have a fallback mechanism in case videos can't be played online for any reason?
4. Are there any other limitations to the current implementation that we should be aware of? Are there any functionalities that still need to be implemented?

dermotduffy:
@ccrisan Appreciate the renewed commitment and clarity. Give me some time (a few days?) to clean up the PR, and get you semi-educated answers to your questions 1-4.

dermotduffy:
Update: I've freshed up the PR, and completed some revised testing that helps answer 1-4. You can find the results and testing environment here:

https://docs.google.com/spreadsheets/d/1-wYvXHxOA-_jFffXt3N8BW2ucqL3SdmfFRY0lmTNS1c/edit?usp=sharing

Conclusion + answers to your questions:

1. There appear to be **significant** compatibility issues with the videos created by more recent motion versions. I haven't been tracking that project, but there appears to have been major work -- and videos that I believe previously played correctly for me when this PR was first created, now no longer do. I will add that this has little to do with this PR AFAIK, but rather the state of browser support for the files created by motion (at least for h264).

2. To be clear, I also know next-to-nothing about HTML5 video -- beyond knowing I want this feature :-) If you have other contacts that are actual experts, and are willing to write the code, you should get their expertise instead. The above spreadsheet has browser compatibility results. The summary is roughly: "Chrome (!Android) yay, everything else nay". This may depend on what external codecs the user has installed, but at least in my fairly vanilla setup other browsers could not play the default h264 output from motion (to be clear: they could play neither the downloaded files using the existing download functionality, nor could they play the content back natively with this PR).

3. The existing "Download" functionality is entirely untouched, and continues to work in all browsers (as per spreadsheet).

4. Yes, there are limitations. As a reminder, in this commit (https://github.com/ccrisan/motioneye/pull/218/commits/b3ea5371d0d0817e4ec9476a7036e3630cf900be) I backed out the support for HTTP Range to simplify this PR. By doing so, seeking through the movie in any browser is not possible. In addition, the browsers thus appear to download the entire movie before allowing playback, which is obviously slower for the user. If you choose to accept this PR, I would recommend we follow it with a subsequent PR that restores the Range support which gives an improved user experience.

4(b): I believe iOS devices (I do not own any, so cannot test) will not playback any video file natively if the server does not support Range. So absence of Range for iOS is not merely an inconvenience as with other browsers, but rather prevents playback entirely. Source: https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/CreatingVideoforSafarioniPhone/CreatingVideoforSafarioniPhone.html#//apple_ref/doc/uid/TP40006514-SW6

I didn't test other codecs, but I can if you want me to test something in particular. Based on the above though, I wouldn't be particularly optimistic about finding anything that "just works" for all browser/OS combinations given the output from motion today. As of this PR, motioneye will not show the play icon if the browser doesn't believe it can play the content based on the mimetype, which does prevent some amount of user frustration. But, in my above tests, the devil appears more likely to be in the codec details ...

I think the options are thus:

 - Don't commit this PR. Users continue to download manually and use external players, etc. Same experience for all.
 - Invest time in first getting motioneye/motion to emit more browser compatible formats -- then pursue this PR.
 - Commit this PR. Some portion of users (in particular non-Chrome users) will be surprised/frustrated when the Play icon does not actually work because their browser cannot play the motion content -- and they have to resort to manually downloading the content.
 - Someone who is an expert in HTML5 video and/or codecs, proposes some other alternatives ...

PaulAnnekov:
> Someone who is an expert in HTML5 video and/or codecs, proposes some other alternatives ...

If I will have a time - I will check. At this time all browsers support mp4 with h.264 codec inside. Range requests support is not required to correctly play the video. If it's missing - seek to non-buffered places won't work. So that's strange why stream is not working in FF/IE.

I think it should work like this:
- canPlayType returns `''`? Hide Play button.
- `error` event on start/during playback? Simply show the error reason and give a link to some page in https://github.com/ccrisan/motioneye/wiki where we will gather some info on how to make video to play correctly.

themouseNZ:
I dont know anything about HTML5 so cant help there but what I have found is if I use the .mp4 container I cannot even get VLC to play the files. I switched to .mkv and it downloads and plays in VLC fine.

dermotduffy:
@PaulAnnekov Thank you for the poke to relook at the results. I now think this codec issue was a red herring on my end. When I delete & recreate the cameras in motioneye -- the resultant files play as originally expected in all browsers. I narrowed it down to some weird values I had in my thread-X.conf files for ffmpeg_variable_bitrate and ffmpeg_bps. If I go with the stock values that motioneye currently uses when it creates those files, everything works as expected. I don't know if motioneye previously set these variables differently, or I hand tweaked those files -- either way, the default configuration works as expected.

I've updated the spreadsheet to show the cross browser performance.

Other comments:
 - The PR already already uses canPlayType as suggested (I can't recall but this may have been your suggestion to begin with!): https://github.com/ccrisan/motioneye/pull/218/files#diff-aea127d9c131be3ccec58f032cf85db8R3508).
 - Contrary to your suggestion that Range request aren't required for playback, I still cannot get Safari to playback without that support (it just shows the error message I have in this PR). This also aligns with the experience of @bobhruska above. It plays fine with that support re-added back in.

ccrisan:
@dermotduffy thanks for the detailed and thorough answers.

Having h264 movies playable on Chrome is already good enough for most users. Having the play button disabled or hidden where the browser thinks the video is not playable may also be enough for a good user experience. In addition, we could have a statically maintained blacklist (browser/platform, movie format/mime type) to improve this decision.

I want to see this feature "in production" soon and I believe we could benefit from other users' feedback and contributions if we push it the way it is right now. I'm going to merge this into dev and I believe we can safely include it as part of the 0.39 release.

As per @PaulAnnekov's recommendation, we could have a wiki page dedicated to troubleshooting in-browser playback issues.

dermotduffy:
Excellent, thanks @ccrisan . Would you like me to prepare a followup PR with the Range support? My tests show this results in an improved user experience (quicker video playing from the start, seeking to arbitrary positions, and (I believe) functionality in Safari).

Let me know.

kristian:
My comment is already half a year back, so... fist of all: Great that there is some new drive to this issue! I really do think that this issue will much benefit all users of motionEye! Thanks to everyone involved! *thumbs up*

Just my 2 cents and providing you another "end user perspective" on the topic again brought up by @dermotduffy: For me (and I think depending on the sensitivity settings etc. of many other users), having range support (so beeing able to skip & scroll through the video), would be an essential feature to have in this! Especially as I think "keeping the code clean" is not a valid statement any longer, as range support wasn't needed in the past and now there is a clear use case for it, so it should be added again in my opinion. The simple reason is, I think many motion captures are triggered due to a lot of different events, sudden sun changes, a car passing by dropping a huge shadow and therefore landing inside of the tracking area etc. etc. In case you really would like to look through the recordings on a daily basis, you essentially will start to just first check the middle of the video, to see if there is something suspicious and if not, delete the video and check the next one. So simply from usability point of view, range support for movie playback would be essential in my pespective. Also having the functionallity work on iOS devices would significantly improve the user base to be able to benefit from this!

themouseNZ:
This is great, once its on Master we can all provide feedback on this. Thanks everyone :-)

ccrisan:
@dermotduffy I think everyone would benefit from having this Range support (now that I also understand what it is about :P). Nevertheless, I have a couple of questions regarding the implementation:

 * do we want/need Range support for all/more than one handlers or is it enough if we have it only for the movie downloading handler?
 * assuming that we have to touch the movie downloading code, are there any downsides or backwards compatibility issues with the good old movie downloading mechanism?

dermotduffy:
@ccrisan Apologies, was traveling. Answers below:

(1a) Ideally we wouldn't have to implement it at all! The ideal solution is for Motioneye to use Tornado's native StaticFileHandler class to serve static content. This handler natively supports Range, rather than us needing to re-implement it. However, this would represent a non-trivial refactor to break the download code out of the existing handler, and override the StaticFileHandler methods to allow us to do fancy stuff that the current download handler does (e.g. remote cameras). That'd definitely be a cleaner, although much more invasive change.

(1b) Short of engaging in that larger refactor, we only need to implement Range support for a single method: the download() method of MovieHandler().

(2) There are no downsides or backwards compatibility issues. Most/all current requests to that handler will just not include a "Range" header, so the current logic will still be used. Even if a request did use the Range header, it'd still work just fine as the browser expects -- and only the data within the requested range would be returned. If a client uses the Range header, they should/do expect to only get a range reply.

Let me know if you'd like me to prepare a PR for this (presumably against dev branch) -- although I wouldn't be able to commit the time for the larger refactor. The smaller point change is pretty much ready to go if you say the word.

dermotduffy:
@ccrisan Wow, when it rains it pours! I just saw https://github.com/ccrisan/motioneye/pull/815/ which includes both a video playback feature (i.e. duplicates the logic from my original PR that you recently merged), but also includes a clean version of Range support without re-implementing it (as described above, it just uses StaticFileHandler).

Suggestion: As @Mrnt has already written their PR, I'd suggest they resolve the conflicts that are very likely caused by the merge of this PR, but then you merge in their PR -- it looks more featureful than my original one, and the Range support is cleaner as it uses StaticFileHandler.

Mrnt:
@dermotduffy thanks for the mention! Your work inspired me to have a crack at it!

I have tried to make it “enhance gracefully” ie it only offers video playback if it is fairly certain a given video can be played, otherwise it continues to offer the preview image as in the current released version so worst case the functionality is no worse than the current release.

I have tested mp4 video playback with chrome on an older android (4.2.2) device as well as chrome, Firefox, safari on macOS 10.13, chrome, IE on windows 8, chrome, safari on iOS 11. But as you know there is an (almost) infinite combination of codecs, browsers and OS versions, so if you care to test it and give feedback that would be appreciated!

ccrisan:
@dermotduffy @Mrnt I've just tested #815 and appears to work fine with my setup. The code also looks pretty good and it's in a state that can be merged into dev.

The embedded Chrome player allows me to skip through the video so I assume that's the Range support that comes with Tornado's StaticFileHandler. What can I say? Nice work people! You just made 10k users happy :)

PR #230, opened 2016-06-17
Digest Authenticate

rolandmoser:
Change custom authentication to digest authenticate


ccrisan:
I must admit this is a lot of work here. A few questions though:

1. Why not a tornado mixin, such as https://github.com/bkjones/curtain?
2. This would completely remove support for signature-based authentication. Shouldn't a proper solution allow choosing between methods?

rolandmoser:
Thanks for having a look. I only realized that the changes were quite extensive after I performed most of them.

Ad 1. I did not find an "of-the-shelf" tornado based digest authentication and as such tried to implement it myself. Reusing existing code is definitely a better idea than a custom implementation. Please let me know if you have a preferred implementation that you would want to use.
Ad 2. Yes, the current changes remove the signature based authentication. In principle I think there are two options: (a) have both implementation co-exist and (b) implement SHA1 with Digest. Option a would be more complicated but backwards compatible. Option b would probably be less complicated but not backwards compatible. Please let me know your preference.

I am definitely interested to get this upstream into your repository. Please let me know your preferences so that I can try to improve the implementation accordingly.

ccrisan:
Ideally we would have a setting in motioneye.conf/settings.py (e.g. `WEB_AUTH_MODE`) with three possible values: `"signature"` (default), `"basic"` and `"digest"`.

I would go with the suggested mixin as it's more compact. Sure it needs some cleanups but their TODO can help. I would take the code as is and add it to the motionEye tree, mentioning the original author in a comment or so.

Then, when the browser is sent a 401 with WWW-Authenticate digest/basic, it would do its thing asking for credentials and handling the authentication phase. I'm not sure if the extra js file that you have added is necessary as the browser supplies the credentials automatically thereafter (am I missing something?).

All the python http clients that you have converted to CURL can be left as they were since they were already using the CURL backend (weren't they?).

Thanks again for your work and yes, let's get this upstream!

ccrisan:
I'm closing this PR in the absence of a feedback. We currently have HTTP basic authentication and signature methods which seem to be fine for most of the use cases.

PR #243, opened 2016-07-05
Add "Date" e-mail header for better mail-client compatibility

luemmel:
Added "Date" e-mail header, because some clients displayed motioneye emails with from date 01.01.1970 because of missing header


ccrisan:
Thanks! I had no idea.

PR #290, opened 2016-08-21
Add "event_end" treatment in the user inteface

### LINKED ISSUE #132: Why no "on_event_stop" action?
I noticed that under "Motion Notification" you can add a command to be executed. However this is only added to the "on_event_start" directive in de thread-?.conf file.
Luckily, if you manually add a "stop" command to "on_event_stop" directive it does not get overwritten.

I use the start / stop commands to flip a virtual motion detection device in Domoticz (a home automation system) whenever motion decides a motion event either starts or stops, so I can use my IP camera as a motiondetector as well.

Why no seperate options to add a mail / webhook / command option  on event start *and* stop?

jcsilva:
These changes allow the user to trigger an action when an event ends. It will be useful when it is needed to treat start and end of an event in different ways.

### DIFF
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -913,6 +913,9 @@ def motion_camera_ui_to_dict(ui, old_config=None):
 
     # event end
     on_event_end = ['%(script)s stop %%t' % {'script': meyectl.find_command('relayevent')}]
+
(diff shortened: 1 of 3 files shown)


ccrisan:
Thanks. I know it's been a year already but better late than never :)

PR #549, opened 2017-06-06
accept zero value motion gap

jasaw:
Allow event gap to be set to zero to run in gapless mode.

Event Gap description from motion config file.

    Event Gap is the seconds of no motion detection that triggers the end of an event.
    An event is defined as a series of motion images taken within a short timeframe.
    Recommended value is 60 seconds (Default). The value -1 is allowed and disables
    events causing all Motion to be written to one single movie file and no pre_capture.
    If set to 0, motion is running in gapless mode. Movies don't have gaps anymore. An
    event ends right after no more motion is detected and post_capture is over.


ccrisan:
This makes sense. Thanks.

Issue #356, opened 2016-11-12
upgrade to new motion - need help ?


an6059:
I am running motioneye frontend on debian jessie  with Motion 3.4.1~rc01 

how do I upgrade to this latest 4.01 version 

simple cli commands would be really helpful 

thank you for any help on this

ccrisan:
I saw you posted this problem on motion's issue tracker as well. Indeed, this has little to do with motionEye, but I'll help you anyway.

Just download the correct deb file from here https://github.com/Motion-Project/motion/releases/tag/release-4.0.1, and run `dpkg -i <yourfile.deb> as root.

retrostrobe:
Installing the deb file caused a number of ffmpeg issues for me. Read somewhere that I had to compile and configure motion against the installed ffmpeg to make it work. Simply clone the motion repository, then enter the line below to configure, then make, and make install (all sudo).

`./configure --without-mysql --without-pgsql --without-sqlite3 --with-ffmpeg=/usr/lib/arm-linux-gnueabihf `

PR #551, opened 2017-06-07
Fixed regex pattern for extracting username from /proc/mounts

lenkawamoto:
The previous pattern truncated a valid username which contained a hyphen (not as the first character) which resulted in MotionEye repeatedly try to unmount a "no longer necessary" mount point which didn't exist.  The end result was that the live preview video switched to the "No Camera" icon, no live preview was showing on the web interface, and no still images nor motion detected movies were being stored in the Network Storage folder.

The previous pattern of "[\w\s]+" (1 or more occurrence of alpha, number, underscore, or any white space) seemed incorrect to me... though I'm not as familiar with non-Linux usernames nor what restrictions exist for usernames in CIFS mounting.  I simply used the default NAME_REGEX pattern specified in /usr/share/adduser/adduser.conf

See github issue #356 (https://github.com/ccrisan/motioneyeos/issues/356) and the follow-up google forums post (https://groups.google.com/forum/#!topic/motioneye/0Cve8obKAX8) for more history on this issue.

### DIFF
--- a/motioneye/smbctl.py
+++ b/motioneye/smbctl.py
@@ -94,7 +94,7 @@ def list_mounts():
             server, share = match.groups()
             share = share.replace('\\040', ' ') # spaces are reported oddly by /proc/mounts
             
-            match = re.search('username=([\w\s]+)', opts)
+            match = re.search('username=([a-z][-\w]*)', opts)
             if match:
                 username = match.group(1)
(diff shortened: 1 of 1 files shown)


ccrisan:
Looks good. Thanks.

PR #553, opened 2017-06-11
Password protection

larisoan:
Admin and User credentials protection with AES. Also added installation specific application UUID


larisoan:
Unexpected effect on camera remove...

PR #555, opened 2017-06-12
Protecting plaintext passwords

larisoan:
Camera connection plaintext passwords are now protected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ run
 dist
 motioneye.egg-info
 dropbox.keys
+build
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -189,9 +189,11 @@ def get_main(as_lines=False):
 
     _get_additional_config(main_config)
     _set_default_motion(main_config, old_config_format=motionctl.has_old_config_format())
-
+  
     _main_config_cache = main_config
-    
+    _main_config_cache['@admin_password'] = decode_password(main_config['@admin_password'].encode('utf-8'))
+    _main_config_cache['@normal_password'] = decode_password(main_config['@normal_password'].encode('utf-8'))
+
     return main_config
 
 
@@ -400,6 +402,7 @@ def get_camera(camera_id, as_lines=False):
         _set_default_motion_camera(camera_id, camera_config)
     
     elif utils.is_remote_camera(camera_config):
+        camera_config['@password'] = decode_password(camera_config['@password'])
         pass
     
     elif utils.is_simple_mjpeg_camera(camera_config):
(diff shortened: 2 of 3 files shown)


ccrisan:
Looks good to me. There are however two problems to be solved before merging:
* The script [`relayevent.sh`](https://github.com/ccrisan/motioneye/blob/master/motioneye/scripts/relayevent.sh) uses the admin password to relay events to motionEye - I guess it wouldn't work with the new encrypted passwords
 * motionEyeOS requires the plain text admin password so that it can set it as unix root password (see [`adminpw`](https://github.com/ccrisan/motioneyeos/blob/master/board/common/overlay/etc/init.d/adminpw)); I believe the simplest solution here would be to make the encryption optional

Does this still work with existing unencrypted configuration files? If users upgrade motionEye, would they be able to still log?

larisoan:
I need to take a deeper look at these questions. 

Anyway, I think motionEye Admin password should be separated from system root password. One good idea is also to add default password for the system.

ccrisan:
We want to keep the unix root password in sync with motionEye's. The reason is that motionEyeOS behaves like a firmware for a camera. Having separate passwords for the various accounts sucks.

The root password (admin is an alias for root on the system) is also used to access the FTP, SMB and SFTP with privileged rights.

larisoan:
About [`relayevent.sh`](https://github.com/ccrisan/motioneye/blob/master/motioneye/scripts/relayevent.sh) file, 

1. I can modify to take password as a parameter or 
2. decrypt password by using a python script. 

Which on do you prefer? 

About motionEyeOS, what choices there are to update root password? How does it work now and can you pinpoint to me place where it is done.

ccrisan:
@larisoan, please excuse my delayed answer. I've been really busy lately.

I have reviewed your PR again and decided to implement a simplified version myself. https://github.com/ccrisan/motioneye/commit/8c88922e7f269d9491ea52878c3382953b863b5f implements this feature.

The reasons why I chose to implement my own version over your PR are:
 * The passwords in your case are still decodable, which in the end defeats the purpose of this PR.
 * You intercepted the password config values at the lowest possible level (`_dict_to_conf()`). I thought it would be better to handle the encryption in `main_ui_to_dict()`.
 * I found your version, in general, unnecessarily complex.

Regarding the motionEyeOS admin password integration, I added support for calling an external program whenever the password changes: https://github.com/ccrisan/motioneye/commit/a713b4afaa7441fc3b03463a2caef933a3de647e. The password is no longer transmitted to the server every time you change a setting.

I really hope you don't mind me rejecting this PR and I thank you for coming up with this idea.

larisoan:
@ccrisan I don't mind at all, I'm happy that you took my idea and decided to implement it. If you don't mind, I can take one off issue and implement few fixes.

PR #593, opened 2017-07-19
Support armv7 (tested on Odroid X with ALARM Arch Linux)

lucabelluccini:
Hello,
I don't know how we should proceed with supporting multiple architectures, but I tested this Dockerfile and seems to work properly.


ccrisan:
Thanks.

PR #631, opened 2017-08-19
Add support for local MMAL cameras

inodlite:
Allows configuration of local MMAL devices in motioneye, which is useful if you want to use any raspistill like options or hack on secondary stream, zero-copy and such.

- Motion has to be compiled with MMAL support enabled (conditional on RPi userland). This requires a separate change in motion configuration. I made a pull request in upstream, but for now I have a patch for motion package in motioneyeos.
- Right now MMAL device selector is conditioned on hasLocalCamSupport in html, but this should be changed to depend on RPi target instead. Not sure how to do it properly.
- Currently, only hardcoded "vc.ril.camera" is available in selector and there is no availability check.

### DIFF
--- a/motioneye/utils.py
+++ b/motioneye/utils.py
@@ -67,6 +67,15 @@ COMMON_RESOLUTIONS = [
     (1920, 1080)
 ]
 
+RPI_MMAL_RESOLUTIONS = [
+    (640, 480),
+    (1280, 720),
+    (1280, 960),
+    (1920, 1080),
+    (1920, 1088),
+    (2592, 1944)
+]
+
 
 def _(x):
     return x  # this could later be replaced by a proper translate function
@@ -336,8 +345,7 @@ def get_disk_usage(path):
 
 def is_local_motion_camera(config):
     """Tells if a camera is managed by the local motion instance."""
-    return bool(config.get('videodevice') or config.get('netcam_url'))
-
+    return bool(config.get('videodevice') or config.get('netcam_url') or config.get('mmalcam_name'))
 
 def is_remote_camera(config):
     """Tells if a camera is managed by a remote motionEye server."""
@@ -348,6 +356,9 @@ def is_v4l2_camera(config):
     """Tells if a camera is a v4l2 device managed by the local motion instance."""
     return bool(config.get('videodevice'))
 
+def is_mmal_camera(config):
+    '''Tells if a camera is mmal device managed by the local motion instance.'''
+    return bool(config.get('mmalcam_name'))
 
 def is_net_camera(config):
     """Tells if a camera is a network camera managed by the local motion instance."""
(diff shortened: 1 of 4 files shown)


ccrisan:
Great work. I have a few questions though:
1 . Do we have any direct performance advantage when using MMAL over acessing the RPi camera via the standard V4L2 interface?
2. Do all other motion functions work when using the MMAL interface?

inodlite:
Thank you. 

1. BM2835 V4L2 driver and MMAL code in motion (as it is now) are basically calling the same MMAL methods. So I don't expect any drastic performance gains, especially on my RPi0. The point here is to have access to all other things that can be done with a lower level API, like passing down some specific CSI params (can be done through extra motion options on mmal devices: mmalcam_control_params "-ex night -awb night", which I don't know how to pass into V4L2) or modding mmal-motion to enable zero copy buffers or secondary stream (as can be seen on), which is hard to achieve without MMAL.
2. My understanding is that, in general, there are no differences between the MMAL or V4L2 in motion. MMAL is simply returning frames as any other source in [https://github.com/Motion-Project/motion/blob/master/video_common.c, under CAMERA_TYPE_MMAL]. I've seen no differences so far and I assume that non-intrusive nature of MMAL changes is why they made into motion at all.

jasaw:
I am particularly interested in the zero copy that you talked about. What's the difference between mmal and v4l2 that enables zero copy on mmal but not v4l2? Can you please explain what is need for zero copy to work? 

Currently, zero copy is forced enabled in ffmpeg and causing motion to hang when used with v4l2. If zero copy works with mmal, that will definitely improve hardware accelerated video encoding.

motioneye/utils.py
```diff
@@ -67,6 +67,15 @@
...
 
+RPI_MMAL_RESOLUTIONS = [
+    (640, 480),
+    (1280, 720),
+    (1280, 960),
+    (1920, 1080),
+    (1920, 1088),
+    (2592, 1944)
+]
+
```
ccrisan: Are we limited to these precise resolutions when using MMAL? Why can't we use any of the other common resolutions?

inodlite:
We’re not limited by those resolutions. Those were the ones that were
defined in RPi camera doc and (except 1080p I think) were known to work
without padding patch.

ccrisan:
@inodlite in that case, do you think it'd be a good idea to simply use [COMMON_RESOLUTIONS](https://github.com/ccrisan/motioneye/blob/master/motioneye/utils.py#L49) instead?

inodlite:
I think it should be fine. Worst case someone picks a resolution that
requires unpadding, which is slower, but it doesn’t really matter as long
as user is happy with resulting FPS.

dcast78:
Hello, some time ago i have installed Motioneye on my RPI3 with raspbian (and other packages) following this wiki page https://github.com/ccrisan/motioneye/wiki/Install-On-Raspbian.
How can I use Motioneye with support for local MMAL cameras?
Can you update instructions on wiki page?
Thanks

PR #687, opened 2017-10-09
Update Dockerfile

magnulu:
added installation of curl, as my motion log file threw errors about not finding curl when taking a manual snapshot.


ccrisan:
Thanks.

PR #760, opened 2018-01-10
Pause camera feed(s) when viewing captured images/movie thumbnails

Mrnt:
When a dialog is present we are most likely reviewing captured images/movies which can cause a lot of thumbnails to be downloaded. If we are capturing and reviewing on a Raspberry Pi 2 or lower then continuing to refresh the camera feed can cause very slow loading of the thumbnails. Since the camera feed(s) are at least partially obscured when the dialog is up, pausing the camera feed seems acceptable.


ccrisan:
Makes sense. Thanks.

PR #785, opened 2018-02-04
Fix mime type for video downloads

Mrnt:
Make videos have mime type specific to the respective format instead of video/mpeg
Addresses:
- https://github.com/ccrisan/motioneye/issues/740
- https://github.com/ccrisan/motioneye/issues/775


ccrisan:
Thanks!

PR #815, opened 2018-03-07
Add video playback support

Mrnt:
- The StaticFileHandler gives us automatic support for mime type and partial file download using the Range request header.
 - Subclass static handler to support video playback and also fo download
 - js changes to preload video and then show video node if video loads
 - progress indicator while video downloading
 - error indicator when video failed to load with error message but display video preview image (as before) instead of browser's broken video icon
 - do not show video on older devices, instead continue to show preview image as before
 - reset width and height properly on every preview load


Mrnt:
@ccrisan @dermotduffy 
I have merged in the changes from Dermot, mostly dropping my JS implementation of video support. However I do have some concerns with the way it stands now:

1) Video from remote cameras has to first be fetched to the Motioneye instance that is serving up the page before it can be made available to the browser, which can take some time. If the user clicks the play button before the movie file has been fetched by the local Motioneye instance, then the browser has to wait for the movie to be fetched from the remote camera and then sent out to the browser which can cause a time out and an erroneous message that the video is not supported by the browser.

2) Relying on the mime type to indicate to the user that the video can be played in the browser is risky. Mime type is a reflection of the video file format but says nothing about the codecs used to encode the video. Motioneye for example, has 3 different options for encoding mp4 video all of which will have the mime type video/mp4 but not all of which will actually be playable in the browser. 

In my original implementation I addressed these two issues by not providing any video playback indicator to the user until enough of the video file was available for the video tag to be checked to see if the video was actually playable. From a user's point of view this would mean waiting for a video play button to appear, but if it did appear there was a high degree of certainty the video would play and if the video was not playable then an error indicator would appear. 

I should note that issue 2 can also be addressed somewhat by passing additional information about the video (and audio) encoding to the browser with the mime type, eg:
`<video src="some_file" type="video/mp4; codecs=avc1">`
would indicate an mp4 file using the H264 video codec (as against HEVC, etc).  When the video tag is polled using canPlayType() it is more likely to give an accurate answer if the codec is supplied. I am not sure how easy it would be to implement this since Motioneye has no way of knowing what codec was used in a particular video file without examining the information stored within the file.

Mrnt:
@kleini do you know why the Travis CI build is failing?

kleini:
Will have a look on the failing build.

kleini:
The login to Docker Hub fails because DOCKER_USERNAME and DOCKER_PASSWORD seem to be missing when building pull requests. I will have to do some research how this is meant to be with GitHub, Travis CI and Docker Hub.

kleini:
According to this documentation https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings DOCKER_USERNAME is not available for forks and their pull request builds.

kleini:
Please see my last commit in pull request #832 how to fix the build.

Mrnt:
OK thanks. Should I go ahead and add that change to the travis file then?

kleini:
@Mrnt you should just cherry-pick my latest commit from my forked branch on top of your branch.

ccrisan:
@Mrnt @kleini the PR is ready to be merged, IMHO. Should I wait till the Docker build issues are fixed or can I merge it right away?

kleini:
Was able to push to @Mrnt fork. Travis build should become green in a moment.

PR #831, opened 2018-03-16
another pull request test

kleini:
with rebased commit to verify if this makes the build not fail


kleini:
every push to docker hub from a pull request coming from a different repository fails due to missing DOCKER_USERNAME and DOCKER_PASSWORD

PR #832, opened 2018-03-16
test pull request

kleini:
which should make the build not fail due to disabled push


ccrisan:
@kleini is this just a test or is it still necessary?

kleini:
not needed any more

Issue #657, opened 2017-09-11
Allow user to override hostname


ioquatix:
My server has several hostnames and the default one is not the one I want for this install.

I manually edited this line: https://github.com/ccrisan/motioneye/blob/306155120776f36140794a1373df8b25ae639692/motioneye/handlers.py#L244

I feel like we should make this a config option. I'd be happy to make a PR if that was acceptable, with the default being the current behaviour unless the configuration is specified.

What do you think?

ccrisan:
I I don't see the point in making the hostname configurable. It doesn't have too much importance in motionEye's functioning.

ioquatix:
I put motion eye on a public IP behind nginx and I don't want it revealing the host name of the server.

ccrisan:
Alright if you can provide a PR to make this configurable while providing backwards compatibility, I'll review and accept it.

Dycast:
I'm running this on my local server and allow a remote desktop approach to reach it when not home, it would help greatly if the ports (ie: :80 ) were also configurable, I believe the seperate cameras are, but not the main one, which I like to use as I can see all cameras on one screen. - Currently I'm running the OS ver, but am looking at using this one laid on top of a distro as I'm expanding and have several RPI's running this around the house, several of which I want to have a 'local monitor' screen attached

ioquatix:
@ccrisan Can you give me a very rough idea of how you'd like this implemented?

ccrisan:
@ioquatix a "@server_name" configuration option would be the best way to implement it. It can be empty (missing) by default in which case the hostname will be used.

PR #845, opened 2018-03-28
add server_name config option

hobbyquaker:
Feature reqeust from #657
I'm not really into python, if this is done horribly bad pls point me in the right direction


hobbyquaker:
oh i guess i should do the pr against your dev branch, will close this one.

PR #846, opened 2018-03-29
add server_name config option

hobbyquaker:
Feature reqeust from #657
I'm not really into python and new to motionEye, if this is done horribly bad pls point me in the right direction.


ccrisan:
Simple and to the point. Thanks.

ccrisan:
By the way, just so you know, we (I) can change the base branch anytime without the need of recreating the PR: )

PR #849, opened 2018-03-29
improve updateLayout

hobbyquaker:
I changed the way `updateLayout()` calculates the pageContainer width on fitFramesVertically. This change makes it work correctly on mobile devices in landscape orientation also.


ccrisan:
Looks good. Thanks!

Issue #914, opened 2018-06-22
Dropbox filename truncation when using symlinks


randomstring:
I have my /var/lib/motioneye directory symlinked to an external USB drive. The path manipulation in motioneye/uploadservices.py is calling os.path.realpath() on the target dir, but not on the filename. When stripping the target_dir from the filename to generate the rel_filename for the dropbox target, if the target_path and the symlink path differ in length, the incorrect number of bytes are stripped off the filename. See the example below where 4 extra characters are removed from the beginning of the target filename.

From the logs:
>  /var/log/syslog:Jun 21 20:31:40 pibplus meyectl[18243]:    DEBUG: executing task "upload_media_file(/var/lib/motioneye/Camera3/pizero0-2018-06-21/20-30-52.mp4)"
>  /var/log/syslog:Jun 21 20:31:41 pibplus meyectl[18243]:    DEBUG: dropbox: uploading file "/mnt/usbdrive/motioneye/Camera3/ro0-2018-06-21/20-30-52.mp4" to dropbox

The fix is to call os.path.realpath() on rel_filename before truncating. I will create a branch and submit a pull request soon.

ccrisan:
Fixed via e9041db457a3555f51bc10b077b20218745cd53b.

PR #915, opened 2018-06-22
Fix for issue #914. dropbox uploader file truncation

randomstring:
Fix incorrect path manipulation when using symlink for dropbox and other upload services. Use os.path.realpath() on rel_filename before truncating for target filename.


ccrisan:
Yep, that makes sense. Thanks.

PR #925, opened 2018-06-27
Url prefix: add support for non-root serving of motion-eye

agordon:
Hello,

I love MotionEye, use it daily on my RPi3.

I needed this feature so that all web activity from the RPi is served through NGINX, and not directly from other ports (e.g. https://rpi/meye/ is motion-eye while https://rpi/foo/ is something else).

Hope you find it useful.
Comments welcomed.
 - gordon

### DIFF
--- a/extra/motioneye.conf.sample
+++ b/extra/motioneye.conf.sample
@@ -21,6 +21,10 @@ listen 0.0.0.0
 # the TCP port to listen on
 port 8765
 
+# URL Prefix to add, or empty string for root URL (the default).
+# e.g. if URL_PREFIX="/foo", motioneye's UI will be at http://0.0.0.0:8765/foo/
+#url_prefix /foo
+
 # path to the motion binary to use (automatically detected if commented)
 #motion_binary /usr/bin/motion
 
--- a/motioneye/handlers.py
+++ b/motioneye/handlers.py
@@ -131,6 +131,24 @@ class BaseHandler(RequestHandler):
         sigs = []
         sigs.append(utils.compute_signature(method, uri,body,password))
         sigs.append(utils.compute_signature(method, uri,body,hashed_pw))
+
+        """Ugly Hack Alert:
+        The javascript code seems to already strip
+        initial directories from the URL, regardless of the URL_PREFIX
+        feature. This happens in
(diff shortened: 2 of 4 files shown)


ccrisan:
And what's wrong with current recommended method of [Running Behind Nginx](https://github.com/ccrisan/motioneye/wiki/Running-Behind-Nginx)?

I use it every day as well and I have it under `/cams`. motionEye's JavaScript automatically detects the path under which it lives and acts accordingly. The rest is done via desired proxy passing.

agordon:
Hello @ccrisan ,

Thank you for your reply.

I must admit I completely missed that wiki page.
Searching for "motioneye" and variations of "non root url" or "url prefix" or "sub directory" or anything 
remotely similar did not reveal that information.

Even that NGINX wiki page only mentions it in passing - now after learning about it I see that motioneye automatically strips any url prefix, so it can be "/cams/" or "/foo/bar/" - it would still work...
And clearly, one can use nginx+motioneye without a sub-url prefix, or a sub url-prefix without nginx - so that two aren't intrinsically connected - it isn't obvious that a wiki page titled "running behind NGINX" explains sub-url-prefixes.

Additionally, this method of silently ignoring the URL prefix only works if motion-eye is behind a front-end proxy, e.g. running https://[ip]:8765/cams/ does not work at all . This is what confused me and lead me to think the feature is not supported - and this behaviour is not documented (or at least I couldn't find it).
Naively, one would think that if a web-app is configured to work with a sub-url-prefix, the prefix works regardless of access method (i.e. with front-end proxy or with IP:PORT). With my patch the URL prefix is explicit, and works with IP:PORT access as well.

However, there's no doubt - now that I'm aware of how motioneye works,
this method "just works" with NGINX and there's no need for my patch.

Thanks for your time,
 - gordon

ccrisan:
@agordon your variant is indeed an explicit one and a valid solution to the problem at hand.

However, the existing version has the advantage of working out-of-the-box, does not require additional settings and can be used for multiple reverse-proxy servers with different configured URLs.

While I really appreciate your work, I believe the existing variant is the most suited for the majority of motionEye users.

Given all these, I will close this PR.

PR #935, opened 2018-07-09
Update config.py

francisuk1989:
Fixed for this error below

[-1976663088:motion] [ALR] [ALL] conf_cmdparse: Deprecated config option "thread" since after version 3.4.1:
[-1976663088:motion] [ALR] [ALL] conf_cmdparse: The "thread" option has been replaced by the "camera" option.

### DIFF
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -38,7 +38,7 @@ import uploadservices
 import utils
 import v4l2ctl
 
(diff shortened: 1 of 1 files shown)


ccrisan:
@francisuk1989 there are two big problems with this PR:

1. Simply changing the file name won't change the directive name in the conf file. So the warning will still be there.
2. We still have motion 3.x users that rely on motionEye. We cannot simply ignore them. Not just yet :)

PR #1033, opened 2018-10-23
Add autocapitalize attribute to usernames

timmo001:
# Description

Sets `autocapitalize="none"` to usernames. This stops mobile devices from automatically capitalising the username input when logging in or setting it in the settings.


ccrisan:
Thanks!

PR #1098, opened 2018-12-25
Change the color of the not active area of the motion mask to red

19grad:
i changed the color of the "not active area" of the motion detection mask from dark to red.

Because i have had the same issue https://github.com/ccrisan/motioneye/issues/589

image


ccrisan:
Sure, I don't see why not. Thanks.

PR #1211, opened 2019-04-02
use motion 4.2.2 for Docker images

kleini:


extra/Dockerfile.armv7-armhf
```diff
@@ -65,6 +66,6 @@ ADD extra/motioneye.conf.sample /usr/share/motioneye/extra/
 
 CMD test -e /etc/motioneye/motioneye.conf || \    
     cp /usr/share/motioneye/extra/motioneye.conf.sample /etc/motioneye/motioneye.conf ; \
-    /usr/local/bin/meyectl startserver -c /etc/motioneye/motioneye.conf
+    /usr/local/bin/meyectl startserver -c /etc/motioneye/motioneye.conf -d
```
ccrisan: Do we really want `-d` here?

kleini: It was necessary for me. Otherwise motioneye told me during network camera registration: "Protocol "rtsp" is not supported by libcurl".

kleini: Maybe I just need more time to investigate, what happened there.

ccrisan:
@kleini I'm planning on releasing 0.40 stable. Do you think we can merge this PR as well, since it's all about Motion 4.2?

bestlibre:
I'm using it without (detected) issues. Not sure if this is really informative since my setup is quite small (2 rtsp cam)

PR #1234, opened 2019-04-13
Making sha-bang more portable

phaethon:
'/usr/bin/env bash' for sha-bang makes the script more portable. It does not break any of the existing supported Linux platform, but now it, also, works on FreeBSD.

### DIFF
--- a/motioneye/scripts/relayevent.sh
+++ b/motioneye/scripts/relayevent.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 if [ -z "$3" ]; then
     echo "Usage: $0 <motioneye.conf> <event> <motion_camera_id> [filename]"


ccrisan:
Where does `bash` live in FreeBSD? As far as I know, shell scripts live in `/bin` by convention.

phaethon:
Sorry, didn't notice your question earlier. `bash` is not part of the core system in FreeBSD, it can normally be found in /usr/local/bin/bash. There are csh, sh, tcsh under /bin. This is why I proposed in the patch using `/usr/bin/env bash` as it should work under most of the *nix configuration I am aware of (at least under FreeBSD and Linux).
Was there any issue with the patch I submitted?

ccrisan:
Honestly I don't know but most of the scripts I've seen do not use `/usr/bin/env` in shebang. Do you have a link where this is the recommended way of doing it?

phaethon:
You can read this: https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my
What it essential does, is getting bash from the path instead of you being responsible for giving the exact path.
Just ran a `grep -irl /usr/bin/env /usr/*` on one Ubuntu system and got 12 scripts under `/usr/bin` using this approach and hundreds of python scripts using it at the start. Of course exact numbers will differ depending on the packages you have installed.

kleini:
I am using this, too, for all my scripts, I am writing on Linux systems. I would agree to change the she-bang.

ccrisan:
Alright, let's have this merged then. Can we also have the change applied to the new `migrateconf.sh` script, for the sake of consistency?

PR #1235, opened 2019-04-14
Add playback speed controls and autoplay next file

bkiziuk:
Adds additional controls to embedded video player:
- Up Arrow => +10 seconds
- Dn Arrow => -10 seconds
- Pg Up => increase playback speed by a factor of 2
- Pg Dn => decrease playback speed by a factor of 2
- Ins => set playback speed to 1
If PgUp or PgDn controls were used, the next file will be played automatically after the current file ends.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ dist
 motioneye.egg-info
 dropbox.keys
 .venv
+venv/
+.vscode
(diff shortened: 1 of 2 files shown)


ccrisan:
Looks good. I'll give it a try. Thanks.

bkiziuk:
no longer valid as it doesn't work so well in 0.41

Issue #1242, opened 2019-04-17
feature request: Provide an option to change user-agent


ttimasdf:
The default UA used by motioneye `Python-urllib/2.7` is banned by either Discord or Cloudflare, directly returned 403. Have the same request sent from httpie works well from the same machine.

The following message is part of the http response from Cloudflare:
```html
            <h2 data-translate="what_happened">What happened?</h2>
            <p>The owner of this website (discordapp.com) has banned your access based on your browser's signature (4c*********a48).</p>
```

PR #1245, opened 2019-04-18
Fix wrong parsing of webhook command

ttimasdf:
This patch partially fix the problem that any custom command containing 'webhook' in the command line will be falsely identified as a webhook command, therefore wrongly displayed in Web UI. Moreover, the config file will be overwritten by the malformed data from UI after clicking Apply.

Example:
I set a command to run after a movie is saved. (send a POST request with httpie)
```
http POST "https://discordapp.com/api/webhooks/my-bot-auth-token" "content=Motion saved file %f"
```
after I saved and refresh the page. the settings page looks like this
image
At this moment, the command will run correctly as `/etc/motioneye/camera-x.conf` remains good.

Then I change anything in web UI and click `Apply` the setting.

The Camera config file goes wrong completely.
```
on_picture_save /home/pi/.virtualenvs/motion/local/lib/python2.7/site-packages/motioneye-0.40rc5-py2.7.egg/motioneye/scripts/relayevent.sh  "/etc/motioneye/motioneye.conf" picture_save %t %f; /home/pi/.virtualenvs/motion/bin/python /home/pi/.virtualenvs/motion/local/lib/python2. 7/site-packages/motioneye-0.40rc5-py2.7.egg/motioneye/meyectl.pyc webhook -c /etc/motioneye/motioneye.conf 'GET' 'content=Motion file at %f'
on_movie_end /home/pi/.virtualenvs/motion/local/lib/python2.7/site-packages/motioneye-0.40rc5-py2.7.egg/motioneye/scripts/relayevent.sh "/  etc/motioneye/motioneye.conf" movie_end %t %f; /home/pi/.virtualenvs/motion/bin/python /home/pi/.virtualenvs/motion/local/lib/python2.7/    site-packages/motioneye-0.40rc5-py2.7.egg/motioneye/meyectl.pyc webhook -c /etc/motioneye/motioneye.conf 'GET' 'content=Motion file at %f'
```

### DIFF
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -1404,7 +1404,7 @@ def motion_camera_dict_to_ui(data):
             except:
                 ui['email_notifications_picture_time_span'] = 0
 
-        elif e.count('webhook'):
+        elif e.count(' webhook '):
             e = shlex.split(e)
 
             if len(e) < 3:
@@ -1448,7 +1448,7 @@ def motion_camera_dict_to_ui(data):
 
     command_storage = []
     for e in on_movie_end:
-        if e.count('webhook'):
+        if e.count(' webhook '):
             e = shlex.split(e)
 
             if len(e) < 3:


ccrisan:
@ttimasdf first off, thanks for the PR - it makes sense and does improve the webhook config detection. However I have some additional remarks:

* While I admit the webhooks configuration UI is not as intuitive as it should, the way you set up your webhooks is not correct: you're supposed to add the request arguments as URL-encoded query arguments:

    https://discordapp.com/api/webhooks/my-bot-auth-token?content=Motion%20saved%20file%20%f

* The same fix should be applied to sendmail as well (I'll take care of it myself).

ttimasdf:
Thanks for your tips and all of your works 👍 The reason of using a custom command is that it allows for some sort of customization, like changing the UA field(#1242) or add other headers for authentication. Meanwhile, I don't have to escape the string or url by myself considering I may use Chinese in the notification string, makes it hard for human. More or less, it works 😄

PR #1247, opened 2019-04-20
Add support for privacy mask

### LINKED ISSUE #1193: Support for new mask option
Hi,

with motion version 4.1 a new mask option has been introduced: mask_privacy (see here: https://motion-project.github.io/motion_config.html#mask_privacy)

The privacy mask works like the previous existing mask option with the difference, that it completely blacks out the marked parts of the camera image for privacy reasons. The detection is done after the privacy mask filter has been applied.

You already stated in issue #797 that changing the current mask editor is too much effort. May I suggest that instead you add an alternate way for providing the mask definition: by either using your current editor, or by specifiying an existing mask file on the server? This option could be used for both mask files.

So users could be able to create the *.pgm mask file by other means and only reference it via these two options (mask_file and mask_privacy)

Keep up the great work!

Regards, Christian

zagrim:
This is my proposal for implementing support for privacy masks in motionEye (fixes #1193). The new Privacy Mask related settings are located in Video Device section using similar controls as with motion detection masks. To avoid the need for migrating config dir contents I wanted to keep the original motion detection mask file naming intact even though otherwise I renamed mask stuff with "motion" or "privacy" to avoid confusing the two.

### DIFF
--- a/motioneye/static/js/main.js
+++ b/motioneye/static/js/main.js
@@ -877,24 +877,22 @@ function initUI() {
         }
     });
     
-    /* disable mask editor when mask gets disabled */
-    $('#maskSwitch').change(function () {
+    /* disable corresponding mask editor when the mask gets disabled */
+    $('#motionMaskSwitch').change(function () {
        if (!this.checked) {
-           disableMaskEdit();
+            disableMaskEdit('motion');
        } 
     });
-    
-    /* disable mask editor when mask gets disabled */
-    $('#maskSwitch').change(function () {
-       if (!this.checked) {
-           disableMaskEdit();
-       } 
-    });
-    
-    /* disable mask editor when mask type is no longer editable */
-    $('#maskTypeSelect').change(function () {
+    $('#privacyMaskSwitch').change(function () {
+        if (!this.checked) {
+             disableMaskEdit('privacy');
+        } 
+     });
+     
+    /* disable motion detection mask editor when mask type is no longer editable */
+    $('#motionMaskTypeSelect').change(function () {
         if ($(this).val() != 'editable') {
-            disableMaskEdit();
+            disableMaskEdit('motion');
         }
     });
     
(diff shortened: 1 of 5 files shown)


zagrim:
Oops, I just noticed I had failed to address some things in the GUI. I'll fix those and submit another PR soon.

ccrisan:
You do realize that you can amend your PR at any time, by pushing more commits to your source branch, don't you? You don't have to create new PRs for each change :)

Anyways, thanks for your contribution.

zagrim:
Some people think it's just polite that if you know your change is flawed and you need to fix it, you don't leave your PR / review open. I knew I had to do some fixing and further testing so I went with that rule :)

motioneye/static/js/main.js
```diff
@@ -1318,23 +1317,34 @@ function enableMaskEdit(cameraId, width, height) {
...
     }
     
     if (!overlayVisible) {
         showCameraOverlay();
     }
 }
 
-function disableMaskEdit(cameraId) {
-    var cameraFrames;
-    if (cameraId) {
```
zagrim: disableMaskEdit never seemed to get called with a cameraId, so I removed related parts from the code.

zagrim:
Ok, now the UI should work as expected.

zagrim:
I've now resolved the conflicts that had emerged. Do you @ccrisan have plans on merging this at some point? Any improvements I should do?

zagrim:
@ccrisan please communicate if you'd be willing to merge this change if I (again) resolved the accumulated conflicts? I probably could do that within few weeks.
It seems like I'm not the only one who'd like to have this feature available.

ccrisan:
Sure, let's see this merged before we carry on with porting to Python 3. I never had the time to actually test this feature, but it looks good to me.

zagrim:
@ccrisan accumulated conflicts have now been resolved, please review and test :)

ccrisan:
@zagrim LGTM.

marcusobrien:
When this "Privacy Mask" front end widget was added, how does the client camera provide these GUI web options to the web hub server doing the rendering in the browser ?

Raspberry Pi Zero W 2
I can see the "privacy mask" when my Raspberry Pi Zero W 2 in the web browser when looking at my motioneyeos hub, I can configure it in "Video Settings"

Raspberry Pi Zero W
But in the Raspberry Pi Zero W with the same motioneyeOS installed as the above, no option appears for "Privacy Mask" under the "Video Settings" in the web GUI

Because there is no Privacy Mask setting on the web front end for the Zero W Pi, whenever I change any settings for that Pi oin the browser, the server and client throgu an error. The client camera Pi Zero W says ERROR 'mask'

File "/usr/lib/python2.7/site-packages/motioneye/config.py", line 940, in motion_camera_ui_to_dict
KeyError: 'mask'

zagrim:
Privacy mask support was merged into the old python2 development branch on Jun 29 2020, and the latest offician release (0.42.1) was done on Jun 6 2020, so it's not available in any of those releases :disappointed: . But I think it is available in any of the pre-releases that have been done with the python 3 support. As far as I know there's not yet been much done with motioneyeOS regarding python3 support, so I can't really see how one could have privacy mask support if using any release of MEOS. Are you sure you're really using MEOS and not standalone MotionEye in that Pi Zero W 2? Also the last part of your comment hints that you've been running quite different versions of MotionEye in Zero W2 vs. Zero W. Anyway, the version of ME you've installed on Zero W2 seems to be quite recent and I wonder if you could use the same installation origin for the Zero W also? (I have absolutely no idea how much those two platforms differ and whether they are even architecturally compatible...)

> File "/usr/lib/python2.7/site-packages/motioneye/config.py", line 940, in motion_camera_ui_to_dict
KeyError: 'mask'

The privacy mask related config option, if you wish to tinker a bit, is [`mask_privacy`](https://motion-project.github.io/motion_config.html#mask_privacy), and it needs to point at a similar image file as is used for the the motion detection mask. For some more details please see my comment here: https://github.com/motioneye-project/motioneye/issues/1193#issuecomment-482639258

PR #1260, opened 2019-05-03
loglevels

ktechies:
loglevels dont seem to affect motion.log

should this not be changed to read this?


ccrisan:
Your changes don't make too much sense to me. `logging.DEBUG` is `10` and `logging.FATAL` is `50`. So  higher log levels should reduce verbosity.

ktechies:
my bad had things mixed up with motion. however changing the loglevel in the config does not seem to affect the actual logs in the motion log file

PR #1307, opened 2019-06-09
Declare when the configuration comes from the cache

hamiltont:
This adds a boolean to the camera configuration when we store it into 
the cache. For all camera configurations, you can easily tell if it is 
the cached configuration or if it is a fresh (non-cached) config based on 
the presence of this boolean

### DIFF
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -421,6 +421,7 @@ def get_camera(camera_id, as_lines=False):
         return None
 
     _camera_config_cache[camera_id] = dict(camera_config)
+    _camera_config_cache[camera_id]['config_from_cache'] = 1
 
     return camera_config
 


ccrisan:
And what is the purpose of this flag? Who/what would the the user that would make use of it?

hamiltont:
May not be worth merging. I've seen/used this before when debugging and trying to determine if the issue is in the core bizlogic or in the cache invalidation. 

In my case, I was debugging why calling the motion API does not result in meye's webUI updating. Thought perhaps it was a caching issue. I'm still not 100% sure, but I'm suspecting motion doesn't have a variable (in the conf file) to represent detection disabled. Calling the motion API therefore only updates the process memory, and since meye's get_camera seems to rely solely on reading motion conf files, it does not know that detection has been disabled using the motion API and so it loads the meye-hack-config of comment lines (e.g `# @motion_detection on`) and presents a UI that is not accurate. (Advice on any of this welcome, if I have the diagnosis right I can fix the issue)

ccrisan:
Motion does not have a config variable for motion detection. It can only be controlled via API. Maybe the solution would be to corroborate motionEye exposed config with what Motion API has to say. But that needs some serious refactorization and in my opinion is not really a priority.

PR #1310, opened 2019-06-13
Clean cloud to handle leading slash properly

bob-lee:


bob-lee:
@ccrisan this is to fix https://github.com/ccrisan/motioneyeos/issues/1924

ccrisan:
Thanks.

PR #1314, opened 2019-06-15
Show slider cursor value on mouse hover

### LINKED ISSUE #1313: Improvement: Display slider value when hovering the cursor
One small thing I've been missing for quite a while in motionEye UI is slider value tooltips. In some cases it would really be helpful for the user if sliders showed their current value (as tooltip, just like they do when adjusting the value) when hovering them. One such case is the motion detection threshold setting, as it has a very fine scale and the cursor position doesn't really tell the exact value. The tooltip would also be nice in other similar cases when there are multiple snap positions between displayed scale values.

I can't think of a case where it would be bad to have such a feature implemented in the sliders, so I went and tried that out myself, and will create a PR soon for it.

zagrim:
Fixes #1313.


ccrisan:
Works well. Thanks!

Issue #1308, opened 2019-06-11
Clean Cloud not working


Saku241:
Hello
I’m using the Google Drive upload service for my motioneye and I enabled the « Clean Cloud » feature. But there’s some problem. At first sight it seems to work : files get removed daily but in reality they dont get deleted they just revoke my ability to watch them and so they still take place in my drive but I cant see them unless I check the largest files (yes my bin is empty)

Can you help me ?
Thanks !

ccrisan:
@Bryan241 can you please attach your logs here?
@bob-lee any ideas?

Bryan241:
> @Bryan241 can you please attach your logs here?
> @bob-lee any ideas?
How can I send you my log, but the problem come from the Google Drive clean cloud feature because on my raspberry pi, folders are deleted correctly and are no longer on my disk

bob-lee:
@Bryan241 your log doesn't show any activity around gdrive cleanup feature. Can you provide longer logs and `camera-1.conf` at `/etc/motioneye` so that I can get some clue? Below is taken from logs of my rpi for example:
>May 24 07:44:23 raspberrypi meyectl[214]:     INFO: cleanup started
May 24 07:44:23 raspberrypi meyectl[214]:     INFO: wsswitch started
May 24 07:44:23 raspberrypi meyectl[214]:     INFO: tasks started
May 24 07:44:23 raspberrypi meyectl[214]:     INFO: mjpg client garbage collector started
May 24 07:44:23 raspberrypi meyectl[214]:     INFO: server started
Jun 02 08:22:56 raspberrypi meyectl[214]:     INFO: gdrive: found 0/2 folder(s) in local/cloud
Jun 02 08:22:56 raspberrypi motion[518]: [1:ml1:Kitchen Doo] [WRN] [ENC] ffmpeg_set_codec: Preferred codec h264_omx has be
Jun 02 08:22:58 raspberrypi meyectl[214]:     INFO: gdrive: deleted a cloud folder '2019-05-24'
Jun 02 08:23:01 raspberrypi meyectl[214]:     INFO: gdrive: deleted a cloud folder '2019-05-23'
Jun 02 08:23:01 raspberrypi meyectl[214]:     INFO: gdrive: deleted 2 cloud folder(s)

ccrisan:
Setting debug logging (in motioneye.conf) could help.

ccrisan:
https://github.com/ccrisan/motioneyeos/issues/1924 may also be related.

[...]

PR #1368, opened 2019-07-14
Fix Google Drive cleanup so it doesn't leave orphaned files behind

chiastic-security:
The current cloud cleanup for Google Drive appears to delete files, but in fact it leaves them orphaned: it removes the old folder as a child of the root, but this doesn't delete its files, or even delete the folder. It makes it invisible, but the files still take up storage space.

To confirm this, look at the web page showing items taking up storage space, and you'll see the supposedly deleted files still there. Before too long, you run out of space entirely.

This PR fixes it by deleting the folder rather than just removing it from its parent.

### DIFF
--- a/motioneye/uploadservices.py
+++ b/motioneye/uploadservices.py
@@ -485,7 +485,7 @@ class GoogleDrive(UploadService, GoogleBase):
             name = self._get_file_title(id)
             self.debug("cloud '%s'" % name)
             to_delete = not exist_in_local(name, local_folders)
-            if to_delete and self._delete_child(folder_id, id):
+            if to_delete and self._delete_file(id):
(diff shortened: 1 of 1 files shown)


bob-lee:
@chiastic-security hi James, I was supposed to use [`files/delete`](https://developers.google.com/drive/api/v2/reference/files/delete) api but actually was using [`children/delete`](https://developers.google.com/drive/api/v2/reference/children/delete) api. This may fix #1308. Does this work well on your side?

chiastic-security:
@bob-lee yes, this PR fixes it for me.

ccrisan:
Alright, let's have it merged than. Thanks!

PR #1369, opened 2019-07-15
Add validation for dirname, filenames and user password to avoid character encoding related issues

### LINKED ISSUE #1363: Having non-ascii in root dir, file names or user password makes motionEye fail
While inspecting and testing things for #1362 I noticed that having non-ascii characters in "Root Directory", "Image File Name" or "Movie File Name" makes things fail. One issue is that os.statvfs() doesn't like to get unicode string as input (and that could be fixed the same way as shlex.split() in the above mentioned PR), another issue is that at least on my dev system (Debian Stretch, en-US.UTF-8 as system locale) ffmpeg calls still failed even though the logged command line was ok when tested outside of the app. My feeling is that it might be hard to get that working correctly in all setups. 
Thus, my proposal would be to just add validator to those fields to prevent users from setting any non-ascii paths/file names.

Along with the same lines, also non-ascii user password seems to fail, using utils.make_str on the input of sha1() didn't fix that fully, so maybe it would also be good to add similar restriction to user password as there already is for admin password, so that it can't contain any non-ascii characters. So I propose having this one also "fixed" by adding validation to user password field in the UI.

zagrim:
This fixes #1363, while I'm not quite sure if filesystem access could have been made to work with non-ascii characters, too.

### DIFF
--- a/motioneye/static/js/frame.js
+++ b/motioneye/static/js/frame.js
@@ -29,7 +29,7 @@ function setupCameraFrame() {
     /* error and load handlers */
     cameraImg.error(function () {
         this.error = true;
-        this.loading = 0;
+        this.loading_count = 0;
         
         cameraImg.addClass('error').removeClass('loading');
         cameraPlaceholder.css('opacity', 1);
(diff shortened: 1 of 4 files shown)


zagrim:
( Uh, I messed up the last rebase beyond my skills of fixing it, so I'll just close this and create a new PR. Sorry.

PR #1488, opened 2019-10-10
API

bforpc:
You have described, to create an API for motioneye.
It will be nice, to have access to the stored Files Directory via the web Browser API, to view / list / ... the stored files.
Otherwise i must install a second web server with everything else on the same machine.

Is there a time plan to realize this?

bfo

### DIFF
--- a/extra/docker-compose.yml
+++ b/extra/docker-compose.yml
@@ -1,17 +1,15 @@
 ---
-version: "2"
+version: "3.5"
 services:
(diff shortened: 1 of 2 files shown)


ccrisan:
I'm sorry but what the heck is this?

bforpc:
Hi Calin,
i'm new with this. Is my question not a pull request?

bfo

PR #1502, opened 2019-10-16
POSIX portable directory and filename validation

jwheeler91:
As discusses in https://github.com/ccrisan/motioneye/issues/1497

### DIFF
--- a/motioneye/static/js/main.js
+++ b/motioneye/static/js/main.js
@@ -17,8 +17,8 @@ var passwordHash = '';
 var basePath = null;
 var signatureRegExp = new RegExp('[^A-Za-z0-9/?_.=&{}\\[\\]":, -]', 'g');
 var deviceNameValidRegExp = new RegExp('^[A-Za-z0-9\-\_\+\ ]+$');
-var filenameValidRegExp = new RegExp('^([A-Za-z0-9/._-]|%[YmdHMSqv])*$');
-var dirnameValidRegExp = new RegExp('^[A-Za-z0-9/._-]*$');
+var filenameValidRegExp = new RegExp('^([A-Za-z0-9 \(\)/._-]|%[YmdHMSqv])+$');
+var dirnameValidRegExp = new RegExp('^[A-Za-z0-9 \(\)/._-]+$');
 var emailValidRegExp = new RegExp('^[A-Za-z0-9_+.%^~<>,-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,4}$');
 var initialConfigFetched = false; /* used to workaround browser extensions that trigger stupid change events */
 var pageContainer = null;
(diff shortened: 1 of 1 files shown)


motioneye/static/js/main.js
```diff
@@ -15,11 +15,11 @@ var resolutionFactor = 1;
...
-var signatureRegExp = new RegExp('[^a-zA-Z0-9/?_.=&{}\\[\\]":, -]', 'g');
-var deviceNameValidRegExp = new RegExp('^[a-z0-9\-\_\+\ ]+$');
-var filenameValidRegExp = new RegExp('^[a-z0-9_/\\%@\(\)-]*$');
-var dirnameValidRegExp = new RegExp('^[a-z0-9_/\\@\(\)-]*$');
-var emailValidRegExp = new RegExp('^[a-z0-9\-\_\+\.\@\^\~\<>, ]+$');
+var signatureRegExp = new RegExp('[^A-Za-z0-9/?_.=&{}\\[\\]":, -]', 'g');
+var deviceNameValidRegExp = new RegExp('^[A-Za-z0-9\-\_\+\ ]+$');
+var filenameValidRegExp = new RegExp('^([A-Za-z0-9/._-]|%[YmdHMSqv])*$');
+var dirnameValidRegExp = new RegExp('^[A-Za-z0-9/._-]*$');
+var emailValidRegExp = new RegExp('^[A-Za-z0-9_+.%^~<>,-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,4}$');
```
zagrim: If not going to allow the syntax `Foo Bar <foo.bar@foosbar.com>` (which I seem to remember being fine for `MAIL FROM` and `RCPT TO` on those servers that I have tested manual SMTP conversations with), maybe it would be better to also exclude `<>` from the address, too, since they can't be part of a valid email address. 

This also doesn't allow TLDs longer than four characters although they seem to be allowed nowadays (see https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#Brand_top-level_domains), and I don't really see any point in being *that* specific with validation :)

jwheeler91: Re: angle brackets - Shall I just add the space back in? Didn't realize I hadn't. I thought I'd copied all the characters into that bit.

Re:  new TLDs -  Oh yeah whoops lol

jwheeler91: Wait that won't work, hang on

zagrim: 👍  I think it's fine for apps like ME to be relaxed about what the user puts in to e.g. email address. Just like in most web forms, the validation is there to make sure the user types in the right kind of stuff. Wrong notification email address won't compromise the application functionality and there's no way of knowing what the email server allows or forbids anyway.

PR #1533, opened 2019-11-07
future: add motion movie_passthrough option

wmig:
https://motion-project.github.io/motion_config.html#movie_passthrough

 When using a RTSP, RTMP, mjpeg and some V4l2 cameras, create movie files of the motion with the packets obtained directly from the camera.


ccrisan:
Thanks!

PR #1588, opened 2020-01-02
Travis CI: trusty is EOL and sudo: is deprecated

cclauss:


ccrisan:
@kleini you seem to have more experience with Travis than I have. Do you see any problem with these proposed changes?

PR #1590, opened 2020-01-02
improved time call accuracy for windows and linux operating systems

bluedistro:


ccrisan:
I'm curious what's the actual difference between the two variants? Does this really improve things on Windows? Does motionEye really run on Windows?

ccrisan:
I'm going to close this PR knowing that:
 * I'm not aware of any accuracy issues or differences between the two calls
 * We're not planning on supporting motionEye on Windows

PR #1591, opened 2020-01-02
Repetition fix

bluedistro:
fixed a seemingly duplicated if statement


ccrisan:
Already dealt with in `python3` branch.

PR #1592, opened 2020-01-02
Range function usability improvement for Python2 and Python3

bluedistro:


ccrisan:
Already dealt with in `python3` branch.

PR #1589, opened 2020-01-02
Travis CI: Use flake8 to find Python 3 compatibility issues

cclauss:
Tests related to #1572

PR #1624, opened 2020-01-26
Travis CI: Use flake8 on the Dev branch

cclauss:
Like #1589 but on the `dev` branch on legacy Python.


ccrisan:
I don't think there's any point in cleaning up the code on legacy code since it's not going to be maintained anymore. Or did I miss something?

PR #1627, opened 2020-01-28
Cycling the videos of the cameras (full screen) - one per time

lozziboy:
Connecting motioneye system to a big LCD screen, will be really useful to have an option that, in monitor view, cycles the videos of the cameras (full screen), one camera per time, and it change the visualized video every 20 seconds.
This is very common in professional security systems and can be a nice feature for MotionEye.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,3 @@
-dist: trusty
-sudo: true
-
 notifications:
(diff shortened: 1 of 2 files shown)


ccrisan:
Please do not use pull requests to create issues.

PR #1709, opened 2020-04-19
Multilanguage

jmichault:
multi-language support.

The language is determined when the server is launched, it does not take into account the preferences of the browser.
By default, it is the language of the system. It can be changed by setting the LANGUAGE variable before launching, for example in the "/etc/systemd/system/motioneye.service" file if motionEye is installed as a service.

All visible texts have been translated into Esperanto in the sources, it is a neutral language well suited as the initial language to be translated.

Translations into the English and French languages are made.

The translations into other languages have been prepared with an automatic translation and therefore require correction by a human before being usable.


zagrim:
I don't know about others and surely I don't have a say in whether to accept this PR or not, but I'd find it very inconvinient to be forced to learn Esperanto to be able to keep contributing to the project (if Esperanto starts to appear in the source files). I know that in theory it would be inpartial and a neutral language and probably even relatively easy to learn for people knowing certain languages, but for the rest it is completely incomprehensible at first sight.

Multi-language support would be neat, and having Esperanto as one of the language options would also be fine, but since the source code already is pretty much English (Python keywords and variables etc) why mix another language there @jmichault, especially as it is not that commonly known language as far as I know?

ccrisan:
@jmichault I'm sorry but I'm not going to accept this PR. Reasons include:
 * as @zagrim pointed out, Esperanto is a no go
 * proper i18n/i10n support should be implemented first in motionEye before proceeding with actual translations
 * there are many tasks that need to be taken care of before translations and I simply don't have the time for them

jmichault:
> I don't know about others and surely I don't have a say in whether to accept this PR or not, but I'd find it very inconvinient to be forced to learn Esperanto to be able to keep contributing to the project (if Esperanto starts to appear in the source files). I know that in theory it would be inpartial and a neutral language and probably even relatively easy to learn for people knowing certain languages, but for the rest it is completely incomprehensible at first sight.
> 
> Multi-language support would be neat, and having Esperanto as one of the language options would also be fine, but since the source code already is pretty much English (Python keywords and variables etc) why mix another language there @jmichault, especially as it is not that commonly known language as far as I know?

Don't have time to learn english.
Learning esperanto took me 10 days.

PR #1741, opened 2020-05-08
Insert py file to sys path first to avoid confusing second instance f…

mmeah:
Ran into an issue when motioneye was installed with pip and also git cloned into a different directory. The service file was modified to use the git directory. The template and static files used kept defaulting to the python directory rather than the originally loaded service path. This was due to the sys.path being appended rather than inserted to the front of the path list.


ccrisan:
Thanks.

PR #1745, opened 2020-05-10
[PY3Experimental] powerctl rewrite (2/10)

Vlsarro:
First thing that caught my eye in this project is that there is a lot of global variables and code duplication. So I wanted this module to be an example for future refactoring for myself. But I was too lazy to continue rewriting of other ctl classes... may be in the future :) If you like this type of code structuring.

Please do not merge until all PRs with "PY3Experimental" prefix pass code review and get approved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+language: python
+
 notifications:
   email:
     recipients:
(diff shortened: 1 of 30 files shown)


Vlsarro:
Messed up pull request creation. Will create later after proper pr chaining :\

PR #1754, opened 2020-05-16
Update main.js

seaniedan:
I specify 'text_event' in 'extra motion options' to create a filepaths for still images and movies based on events. This useful functionality appears to been removed recently: I get 'special characters are not allowed in file name' and cannot set the value.

filenameValidRegExp should allow Conversion Specifier '%C' so that images and movies can reference the value defined by 'text_event'.

Perhaps other Conversion Specifiers (https://motion-project.github.io/motion_config.html) should be allowed too, but I definitely need this one!


ccrisan:
Makes sense. Thanks!

Issue #1785, opened 2020-06-06
MotionEYE needs a webpage, changelog and update methods


starforcee:
I have now 2 separate setups:

One natively on a Raspberry PI3:

motionEye Version 	0.42
Motion Version 	4.2.2+gitUNKNOWN
OS Version 	motionEyeOS dev20200127

One in docker ccrisan/motioneye   master-armhf    (from this I couldn't even tell what version is this because that whole section is missing from the menu along with the UPDATE function).

The prior says that is the latest version. All in all I see changes in the github but there isn't even a proper changelog for the project so one can decide if the recent changes worth an upgrade or not.

Please also consider making a normal webpage instead of using github as the front page for this project.

Thank you.

zagrim:
For the dockerized one, have you tried logging in as "admin" user? (I have not used the docker builds myself, so just guessing here).

There is a changelog for the project: https://github.com/ccrisan/motioneye/releases
That's rather standard place for that kind of info for Github projects, but I agree that for people not familiar with Github that's really hard to find. Maybe the top level README.md should contain a link to that page, too (and maybe some other helpful info as well?). Maybe @ccrisan (or someone else with possibility to do that) could add that?

I think no other varieties of MotionEye have a self-update except for MotionEyeOS (and seems like that's [somewhat limited, too](https://github.com/ccrisan/motioneyeos/wiki/FAQ#how-can-i-update-my-motioneyeos-setup-to-the-latest-version)). Upgrading installations made with `pip` should be easy enough usually, anyway, there are instructions for that in the Wiki top level installation page (and link to the Wiki is present on the project front page), and I would assume the docker builds are upgraded like any other dockerized software, that is by doing again `docker pull` etc, but as I said, I don't have any actual experience with ME on docker.

PR #1787, opened 2020-06-07
Add link to release notes in README

zagrim:
Based on #1785 added a link to release notes to make it easier for people not familiar with Github (and with no interest in the "code" section) to find that information. Maybe the same link should be added to the Wiki, too, since every other bit of information is already there).


ccrisan:
Makes sense. Thanks.

PR #1822, opened 2020-06-30
Fix docker build, move to Debian buster.

### LINKED ISSUE #1818: Dockerfile does not build
The existing Dockerfile won't build a new image since it's based on ubuntu 18:10, which is not a LTS version and thus unavailable in hub.docker.org.

I cleaned up the existing Dockerfile and converted it to use debian buster (with a locked version of SID for the components that need SID). Please let me know if you're accepting contributions and I'll end the PR.

marcopaganini:
- Moved to Debian buster as ubuntu 18.10 is no longer available in docker hub.
- Some packages need to come from sid (E.g. motion itself). In that
  case, make the build reproducible by fixing sid to a snapshot.

Fixes #1818

### DIFF
--- a/extra/Dockerfile
+++ b/extra/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.10
+FROM debian:buster-slim
 LABEL maintainer="Marcus Klein <himself@kleini.org>"
 
 ARG BUILD_DATE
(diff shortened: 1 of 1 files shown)


kleini:
Somehow the Travis CI build fails. I get emails about it, but it is not shown here in the PR. Strange.

marcopaganini:
This is interesting. I was looking at the usual links to Travis here but couldn't find any!

kleini:
I see failing builds here: https://travis-ci.org/github/marcopaganini/motioneye/builds

marcopaganini:
Yes, but those appear to be because it's trying to use a non-existing
environment var:

```
The command "docker build --build-arg VCS_REF=$TRAVIS_COMMIT --build-arg
BUILD_DATE=$(date +"%Y-%m-%dT%H:%M:%SZ") -t
$DOCKER_REPO:$TRAVIS_BRANCH-$TARGET -f extra/Dockerfile${EXT} ." exited
with 125.
0.06s$ docker run --rm $DOCKER_REPO:$TRAVIS_BRANCH-$TARGET uname -a
docker: invalid reference format.
See 'docker run --help'.
The command "docker run --rm $DOCKER_REPO:$TRAVIS_BRANCH-$TARGET uname -a"
exited with 125.
0.05s$ if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then echo $DOCKER_PASSWORD
| docker login -u "$DOCKER_USERNAME" --password-stdin; fi
Must provide --username with --password-stdin
```

Perhaps because some of these (like username/password) are set for the
"official" dockerhub account? They don't seem related to my PR (but I'm
unsure).

marcopaganini:
Yeah, looking again, it appears DOCKER_REPO, DOCKER_USERNAME, and DOCKER_PASSWORD are unset  in my repo. This makes sense. I'm assuming you set those in your private travis account so it can connect to the official dockerhub for the project.

marcopaganini:
So, I took a more careful look at the `.travis.yml` file on the repo.

There are two "phases" of testing a Dockerfile change:
1.  Build it locally (executed for all cases)
2.  Push it to the docker hub, which is only executed on the original repo (yours) and if the request is a merge, *not* a Pull Request (for obvious reasons).

One problem is that the `DOCKER_REPO` variable should be defined in the `.travis.yml` file since it's not a secret of any kind (it's not). It is apparently defined in your repo settings on github, which I can't see since I'm not an owner. Any reason why that variable is not defined in `.travis.yml`? If the idea is to use a repo settings, then maybe we should warn contributors to set their forked repos as well, or their CIs will always fail.

I hacked a travis config in another branch (by hardcoding the DOCKER_REPO) and the test appears to work fine.

Also I can't see the travis run for the PR in your repo. Can you tell me what failed?

kleini:
I am just another contributor, therefore I can not say, why the build does not run in this PR. As already said, maybe PR is created in a wrong way.

marcopaganini:
Hmmm, I'm curious myself. The PR seems innocent and if I remove the docker hub stuff from `.travis.yml`. In fact, I'm somewhat sure that if we added the DOCKER_REPO in the environment section of the `.travis.yml` file, it would work. Is this working today or something that hasn't been changed in a long time? Would could copy me the log messages in the upstream (ccrisan) repo for this run of Travis?

kleini:
Look at all the other PRs. Running Travis stopped around April somehow.

marcopaganini:
Can I add one line to the `.travis.yml` file to test my theory?

kleini:
https://travis-ci.org/github/ccrisan/motioneye/builds and dev is broken, too. So Calin needs to fix that.

kleini:
You can use your PR to change whatever part of the whole code.

marcopaganini:
Thanks. Even though I didn't change the  .travis.yml file, it appears
Travis is actually passing for my PR:
https://travis-ci.org/github/ccrisan/motioneye/builds/703649245

ccrisan:
Docker builts are broken for some time now, indeed. I'm open to suggestions on how to fix them.

marcopaganini:
I believe the problem is that the existing Dockerfile is trying to download
Ubuntu 18.10, which is now deprecated everywhere. I have a pending PR (
https://github.com/ccrisan/motioneye/pull/1822) which moves the Dockerfile
creation to Debian using snapshots, so it should fix this issue.

ccrisan:
Alright, let's see if it works now.

PR #1828, opened 2020-07-04
Delete also thumbnails with movies

### LINKED ISSUE #1669: thumbnails not removed.
After deleting video files all thumbnails are left behind

zagrim:
Closes #1669. Used the same logic which is being used for media cleanup.


ccrisan:
Thanks.

PR #1914, opened 2020-09-26
add webhook for motion event ends

Links2004:
no more motion detected webhook

see #1729

### DIFF
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -1185,12 +1185,16 @@ def motion_camera_ui_to_dict(ui, prev_config=None):
         on_event_end.append(line)
 
     if ui['web_hook_end_notifications_enabled']:
-        url = re.sub('\\s', '+', ui['web_hook_end_notifications_url'])
+        url = re.sub(r'\s', '+', ui['web_hook_end_notifications_url'])
 
-        on_event_end.append("%(script)s '%(method)s' '%(url)s'" % {
-            'script': meyectl.find_command('webhook'),
-            'method': ui['web_hook_end_notifications_http_method'],
-            'url': url})
+        on_event_end.append(
+            "%(script)s '%(method)s' '%(url)s'"
+            % {
+                'script': meyectl.find_command('webhook'),
+                'method': ui['web_hook_end_notifications_http_method'],
+                'url': url,
+            }
+        )
 
     if ui['command_end_notifications_enabled']:
         on_event_end += utils.split_semicolon(ui['command_end_notifications_exec'])
@@ -1625,7 +1629,7 @@ def motion_camera_dict_to_ui(data):
     ui['telegram_notifications_picture_time_span'] = 0
     command_notifications = []
     for e in on_event_start:
-        if e.count(' sendmail '):
+        if ' sendmail ' in e:
             e = shlex.split(
                 utils.make_str(e)
             )  # poor shlex can't deal with unicode properly
@@ -1653,7 +1657,7 @@ def motion_camera_dict_to_ui(data):
             except:
                 ui['email_notifications_picture_time_span'] = 0
 
-        elif e.count(' sendtelegram '):
+        elif ' sendtelegram ' in e:
             e = shlex.split(
                 utils.make_str(e)
             )  # poor shlex can't deal with unicode properly
@@ -1670,7 +1674,7 @@ def motion_camera_dict_to_ui(data):
             except:
                 ui['telegram_notifications_picture_time_span'] = 0
 
-        elif e.count(' webhook '):
+        elif ' webhook ' in e:
             e = shlex.split(
                 utils.make_str(e)
             )  # poor shlex can't deal with unicode properly
@@ -1682,7 +1686,7 @@ def motion_camera_dict_to_ui(data):
             ui['web_hook_notifications_http_method'] = e[-2]
             ui['web_hook_notifications_url'] = e[-1]
 
-        elif e.count('relayevent'):
+        elif 'relayevent' in e:
             continue  # ignore internal relay script
 
         else:  # custom command
@@ -1699,8 +1703,10 @@ def motion_camera_dict_to_ui(data):
 
     command_end_notifications = []
     for e in on_event_end:
-        if e.count(' webhook '):
-            e = shlex.split(utils.make_str(e)) # poor shlex can't deal with unicode properly
+        if ' webhook ' in e:
+            e = shlex.split(
+                utils.make_str(e)
+            )  # poor shlex can't deal with unicode properly
 
             if len(e) < 3:
                 continue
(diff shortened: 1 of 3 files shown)


MichaIng:
Many thanks for your PR. Generally a good idea, so one may get some notification that there is no motion anymore and hence no need to have a look at the camera. At least this is my idea on it 😄.

@Links2004 
Since there are conflicts, could you rebase onto current `dev` and resolve them?

motioneye/config.py
```diff
@@ -1007,6 +1007,14 @@ def motion_camera_ui_to_dict(ui, prev_config=None):
     # event end
     on_event_end = ['%(script)s stop %%t' % {'script': meyectl.find_command('relayevent')}]
 
+    if ui['web_hook_end_notifications_enabled']:
+        url = re.sub('\\s', '+', ui['web_hook_end_notifications_url'])
```
cclauss, line 1011:
```suggestion
        url = re.sub(r'\s', '+', ui['web_hook_end_notifications_url'])
```

motioneye/config.py
```diff
@@ -1453,7 +1462,17 @@ def motion_camera_dict_to_ui(data):
 
     command_end_notifications = []
     for e in on_event_end:
-        if e.count('relayevent') or e.count('eventrelay.py'):
+        if e.count(' webhook '):
```
cclauss, line 1465:
```suggestion
        if ' webhook ' in e:
```

motioneye/config.py
```diff
@@ -1453,7 +1462,17 @@ def motion_camera_dict_to_ui(data):
...
+            e = shlex.split(utils.make_str(e)) # poor shlex can't deal with unicode properly
+
+            if len(e) < 3:
+                continue
+
+            ui['web_hook_end_notifications_enabled'] = True
+            ui['web_hook_end_notifications_http_method'] = e[-2]
+            ui['web_hook_end_notifications_url'] = e[-1]
+
+        elif e.count('relayevent') or e.count('eventrelay.py'):
```
cclauss, line 1475:
```suggestion
        elif 'relayevent' in e or 'eventrelay.py' in e:
```

MichaIng:
Rebased and annotations resolved. Ready for review and testing.

motioneye/config.py
```diff
@@ -1690,7 +1703,19 @@ def motion_camera_dict_to_ui(data):
 
     command_end_notifications = []
     for e in on_event_end:
-        if e.count('relayevent') or e.count('eventrelay.py'):
+        if ' webhook ' in e:
+            e = shlex.split(
+                utils.make_str(e)
+            )  # poor shlex can't deal with unicode properly
```
cclauss, line 1709: Move the comment to its own line to avoid the three-line wrap...
```suggestion
            # poor shlex can't deal with unicode properly
            e = shlex.split(utils.make_str(e))
```
My bet is that Python 3 code has less trouble with Unicode because all `str` are Unicode utf-8 by default.

MichaIng: I don't even know yet what `shlex` actually is and this is used for 😄. But your suggestion makes sense. #2351 added a few more such cases.

cclauss: https://docs.python.org/3/library/shlex.html

MichaIng: You mean the following would work as well?
```py
e = shlex.split(e)
```

cclauss: If `e = Elephant()` or e = `b`bytes`` then no, it won't work but `dev` has no `utils`.
https://github.com/motioneye-project/motioneye/blob/dev/motioneye/utils.py

MarcelCoding (approved):

motioneye/static/js/main.js
```diff
@@ -646,14 +646,21 @@ function initUI() {
...
         return true;
     }, '');
     makeCustomValidator($('#webHookNotificationsUrlEntry'), function (value) {
         if (!value.match(webHookUrlValidRegExp)) {
-            return "use of semicolon (;) or single quote (\') is not allowed in web hook URL";
+            return "use of semicolon (;) or single quote (') is not allowed in web hook URL";
+        }
+
+        return true;
+    }, '');
```
MichaIng, line 660: This block doubles with the one above, or do I need glasses? 🤓

motioneye/templates/main.html
```diff
@@ -1134,6 +1134,31 @@
                         <tr class="settings-item" depends="motionDetectionEnabled">
                             <td colspan="100"><div class="settings-item-separator"></div></td>
                         </tr>
+                        <tr class="settings-item" depends="motionDetectionEnabled">
+                            <td class="settings-item-label"><span class="settings-item-label">Call A End Web Hook</span></td>
```
MichaIng, line 1138:
```suggestion
                            <td class="settings-item-label"><span class="settings-item-label">End Event Web Hook</span></td>
```

motioneye/templates/main.html
```diff
@@ -1134,6 +1134,31 @@
...
+                            <td class="settings-item-value"><input type="checkbox" class="styled notifications camera-config" id="webHookEndNotificationsEnabledSwitch"></td>
+                            <td><span class="help-mark" title="enable this if you want a URL to be requested whenever a motion event ends">?</span></td>
+                        </tr>
+                        <tr class="settings-item" required="true" depends="webHookEndNotificationsEnabled motionDetectionEnabled" strip="true">
+                            <td class="settings-item-label"><span class="settings-item-label">Web Hook URL</span></td>
+                            <td class="settings-item-value"><input type="text" class="styled notifications camera-config" id="webHookEndNotificationsUrlEntry" placeholder="e.g. http://example.com/notify/"></td>
+                            <td><span class="help-mark" title="a single URL to be requested when motion event ends the following special tokens are accepted: %Y = year, %m = month, %d = day, %H = hour, %M = minute, %S = second, %q = frame number, %v = event number">?</span></td>
```
MichaIng, line 1145:
```suggestion
                            <td><span class="help-mark" title="The single URL to be requested when motion a event ends. The following special tokens are accepted: %Y = year, %m = month, %d = day, %H = hour, %M = minute, %S = second, %q = frame number, %v = event number">?</span></td>
```

motioneye/config.py
MichaIng: Works well indeed without `utils.make_str` with various web hooks and commands tested.

MichaIng:
Added headings to the notification section to clearer separate between motion start and motion end actions.
notification

MichaIng (approved):

MichaIng:
Merging now. Tested everything up and down, also the changes I did like merging the URL validation, skipping `make_str(e)` when splitting the event arguments via `shlex.split()`. Text polishing along with translations can be done in a dedicated PR.

PR #1936, opened 2020-10-26
Update to uploadservices.py to reduce the Google Drive scope to drive…

moonlitknight:
….file

At line 329, motionEye is requesting a Google Drive scope of "drive" which grants access to *all* of a user's files. This PR reduces the scope to "drive.file" which allows motionEye to create files on a user's Google Drive and to then only access/delete those files. This is the recommended best practice from Google, ie. to only request the minimum scope necessary for the application's needs. See https://developers.google.com/drive/api/v2/about-auth


ccrisan:
Thanks!

PR #1975, opened 2020-12-23
Add TimeLaps Video Play

Lu-Fi:
Added TimeLapse Functionality by adding a second Button, with this button play speed is set to 5 and next video will start automatically until last video was palyed.
If possible please add a FF symbol to this second Play Button.


jmichault:
Exactly what I wanted most for motioneye. I integrated it into my own [fork.](https://github.com/jmichault/motioneye.eo)
Thank you so much.

PR #1982, opened 2020-12-29
docker compose override with reasonable defaults

itsb:
Some critical additions: timezone sync via /etc/localtime volume mount, and device visibility via device option. Also using the restart option for service persistence and pointing to the dev image.


ccrisan:
Thanks!

PR #1983, opened 2020-12-29
Compatible with python3.7 and later

kilisima:
Three corrections have been made.
(1) Type conversion process in the Lamda function.
(2) Modified to open MediaFile as Binary type.
(3) Changed the data argument of urllib.request from Str type to Bytes type.

Please confirm.


ccrisan:
Looks alright.

PR #2361, opened 2022-03-15
Added GitHub actions docker workflow

MarcelCoding:


MichaIng:
Btw, those shouldn't be required:
- default-libmysqlclient-dev
- libmicrohttpd12
- libpq5
- mosquitto-clients

Those are `motion` dependencies and pull in by the motion package automatically.

extra/Dockerfile
```diff
@@ -1,34 +1,19 @@
...
 ARG RUN_UID=0
 ARG RUN_GID=0
 
 COPY . /tmp/motioneye
 
 RUN echo "deb http://snapshot.debian.org/archive/debian/$(date +%Y%m%d) bullseye main contrib non-free" >>/etc/apt/sources.list && \
-    chmod o+r /etc/resolv.conf && \
+    #chmod o+r /etc/resolv.conf && \
     apt-get update && \
     DEBIAN_FRONTEND="noninteractive" apt-get -t bullseye --yes --option Dpkg::Options::="--force-confnew" --no-install-recommends install \
```
MichaIng, line 13:
```suggestion
    DEBIAN_FRONTEND="noninteractive" apt-get --yes --option Dpkg::Options::="--force-confnew" --no-install-recommends install \
```
Should be redundant since this is a Bullseye system already. Also there may be newer versions available in `bullseye-updates` or `bullseye-security` suites, which we override by this, AFAIK.

MarcelCoding:
@MichaIng what is the branch from which the `edge` docker tag should be built? (`edge` is like the current development version and the preview for the next release.)

extra/Dockerfile
```diff
@@ -50,7 +36,7 @@ RUN echo "deb http://snapshot.debian.org/archive/debian/$(date +%Y%m%d) bullseye
     pip install /tmp/motioneye && \
     # Cleanup
     rm -rf motioneye && \
-    apt-get purge --yes python3-setuptools python3-wheel && \
+    apt-get purge --yes python3-setuptools python3-wheel python3-pip && \
     apt-get autoremove --yes && \
```
MichaIng, line 40:
```suggestion
    apt-get --yes autopurge && \
```
This also removes the config files of removed packages.

extra/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+test -e /etc/motioneye/motioneye.conf || \
+cp /usr/share/motioneye/extra/motioneye.conf.sample /etc/motioneye/motioneye.conf ; \
```
MichaIng, line 3: This does not work anymore. Shipping files like this is deprecated with setuptools, they are now shipped as part of the Python package itself. Hmm, I think best is to use the new `motioneye_init` command here as well. It also includes all non-Python dependencies, setting up default config and systemd unit. Okay the letter is not required in Docker, and not available, I guess, so we'd need to add a CLI option to skip systemd setup.

MichaIng:
See the setup script made available here: https://github.com/motioneye-project/motioneye/blob/dev/motioneye/extra/linux_init
This is invoked by `motioneye_init` command. We should use it here as well and remove the included steps from the Dockerfile + entrypoing.

But we need to add a `--skip-systemd` option, and if passed, exit the script before the last 4 lines.

MarcelCoding:
I am going to refactor the dockerfile as a whole in the next step, I wouldn't recommend executing this script, sometimes it is recommended to do things differently in docker than outside of docker.

MichaIng:
You mean this for uploads to Docker hub, right? Neither do I have a Docker hub account, nor do I have much experience with Docker, so let's leave this to be a container build test only, for now, until we feel ready for a beta.

But in general, I guess we want to update this container: https://hub.docker.com/r/ccrisan/motioneye
@ccrisan not sure whether it's possible to transfer a Docker container or rename an account the same way it can be done with a GitHub repository, so that `docker pull ccrisan/motioneye` will pull from e.g. `motioneye-project/motioneye` automatically, so that we can take over in a way that users don't need to change their scripts/commands/behaviour?

MarcelCoding:
Currently, the images are published to the GitHub container registry. https://github.com/motioneye-project/motioneye/pull/2361/files#diff-17e3400d876978d12bbad517dcf82312b54fa62723408b1ce011b2755458bdafR28 The cool thing is that you don't have to configure any secrets or so any. But we can easily add more docker repos.

MichaIng:
@MarcelCoding 
Any idea how to achieve such a conditional in Dockerfile?
```sh
if [[ $(uname -m) == 'armv7l' ]]
then
    packages=()
    echo -e '[global]\nextra-index-url=https://www.piwheels.org/simple/' > /etc/pip.conf
else
    packages=('python3-dev' 'libcurl4-openssl-dev' 'gcc' 'libssl-dev')
fi
...
... python3 python3-pip "${packages[@]}"
```
So if this is an ARMv7 container (I think `uname -m` is already wrong then), add piwheels.org as source for pre-compiled Python wheels, which includes `pycurl`, hence no build is required and the development headers can be skipped. Also this is bash syntax with the conditional and array, of course, not sure what shell Dockerfile uses and what syntax can be used?

MarcelCoding:
@MichaIng https://github.com/MarcelCoding/onedrive-docker/blob/main/Dockerfile.debian#L15-L31 there is an example. 

I intentionally pulled all pip dependencies as Debian packages, because there are mostly already compiled (for a wide range of architectures) and therefore don't need build dependencies that we have to maintain and it reduces the build time extremely.

MichaIng:
The only package that needs to be build is `pycurl`, on ARMv7 with piwheels.org none at all. An argument for pip is that we get latest upstream versions, while Debian ships quite outdated versions. Not sure whether a few minutes longer is an issue, since we do not need to stare at the GitHub actions to finish? 🙂

MichaIng:
```
W: Conflicting distribution: https://packages.microsoft.com/ubuntu/20.04/prod focal InRelease (expected focal but got testing)
E: Repository 'https://packages.microsoft.com/ubuntu/20.04/prod focal InRelease' changed its 'Origin' value from 'microsoft-ubuntu-focal-prod focal' to 'microsoft-ubuntu-focal-prod testing'
E: Repository 'https://packages.microsoft.com/ubuntu/20.04/prod focal InRelease' changed its 'Label' value from 'microsoft-ubuntu-focal-prod focal' to 'microsoft-ubuntu-focal-prod testing'
E: Repository 'https://packages.microsoft.com/ubuntu/20.04/prod focal InRelease' changed its 'Codename' value from 'focal' to 'testing'
```
Is GitHub in the process of switching the runners to Ubuntu Jammy 🤔.

extra/Dockerfile
```diff
@@ -1,72 +1,36 @@
...
+      python3 python3-pip $PACKAGES tzdata && \
+    # Change uid/gid of user/group motion to match our desired IDs. This will
     # make it easier to use execute motion as our desired user later.
-    sed -i -e "s/^\(motion:[^:]*\):[0-9]*:[0-9]*:\(.*\)/\1:${RUN_UID}:${RUN_GID}:\2/" /etc/passwd && \
-    sed -i -e "s/^\(motion:[^:]*\):[0-9]*:\(.*\)/\1:${RUN_GID}:\2/" /etc/group && \
-    pip install /tmp/motioneye && \
+    sed -i "s/^\(motion:[^:]*\):[0-9]*:[0-9]*:\(.*\)/\1:${RUN_UID}:${RUN_GID}:\2/" /etc/passwd && \
+    sed -i "s/^\(motion:[^:]*\):[0-9]*:\(.*\)/\1:${RUN_GID}:\2/" /etc/group && \
+    pip3 install /tmp/motioneye && \
+    motioneye_init --skip-systemd --skip-apt-update && \
```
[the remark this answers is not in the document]
MarcelCoding: No, because the copy is executed at build time and the volume is created at runtime.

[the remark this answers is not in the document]
MichaIng: Okay, then we need to copy it to another fixed path, and restore it from there in entrypoint. I'll do that.

.github/workflows/docker.yml
```diff
@@ -2,7 +2,6 @@ name: Docker
 
 on:
   push:
-    branches: [ main ]
```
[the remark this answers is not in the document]
MichaIng: The aim was to run this workflow as a test on every push for now, until development has settled for a release. It could make sense to split of into two workflows then, one as a test which runs on every push/PR, one for publishing the image, which runs only when pushing to `master` (we should rename it to `main` indeed). What do you think?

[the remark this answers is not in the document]
MichaIng: Ah lol, without defining branches, if tags are defined, it seems to not run on any push.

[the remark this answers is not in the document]
MarcelCoding: you have to do a * in branches or remove tags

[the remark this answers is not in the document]
MichaIng: Understood the conditions for registry and image upload now and changed them to happen only when a push is done to our `main` branch. So in every other situation is runs as test only.

.github/workflows/docker.yml
```diff
@@ -53,7 +54,7 @@ jobs:
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v1
-        if: github.event_name != 'pull_request'
+        if: github.event_name == 'push' && github.repository.full_name == 'motioneye-project/motioneye' && github.base_ref == 'refs/heads/main'
```
MarcelCoding, line 57: Now tags aren't pushed anymore

MichaIng: Only push events to our `main` branch, which is what we want?

MarcelCoding: And push events to tags.

MichaIng: Ah, this is how it works and the reason it failed before with the "tags required" error message. Then it makes more sense to check for `steps.meta.outputs.tags` instead of `main` branch. Or is there a way to check for both: That a tag exists and the it is associated with the `main` branch?

MarcelCoding: Tags require does it says because now the event is a branch and not pull request anymore. It is configured to create a tag on the default branch with `edge`, on a pr `pr-<id>` and on a tag `x`, `x.x` and `x.x.x.`. But on an push that is not to the default branch, it does not know what tag it should create. Your last idea would be a solution.

MichaIng: Ah got that, `docker/metadata-action` creates the tag meant by the error message. So what is listed below `tags` includes the condition, like that it is a PR `type=ref,event=pr`?

Generally I'd like to keep the workflow running on every push event, so that one can do pushes, also on forks, locally to any feature branch, review the checks, including whether the Docker build still works, before opening a PR. But of course Docker Hub related steps must be skipped. I'm not sure which condition works best and can be achieved easily either for explicit releases only and optionally for pushes to a testing branch => `edge` builds.

I'll not touch the Docker workflow anymore now, since I obviously lack the understanding 😅. If the above aim is not easy to achieve, feel free to revert the events/conditions. While I like tests to run on every push, this is not mandatory when it gets too complicated. The build/install/setup steps looks fine now, including some extensions to `motioneye_init` for scripts and workflows which make sense anyway.

MichaIng: > Is that correct?

Sounds good. The `edge` branch does not exist yet, and probably we want to call it differently, but this is trivial to adjust then. Generally having some sorts of testing branch which is not `dev` (too frequently pushed to) sounds very reasonable.

MichaIng: Okay, actually you are right. However, it shouldn't be `dev` yet, until we found out whether the `ccrisan/motioneye` namespace transfer/redirect is possible or whether we need to create an own account, hence we have Docker Hub credentials.

MichaIng:
Btw, do you see anything useful we may want to adapt form #1948? The main feature to choose a `motion` version, would be better suited in [`linux_init`](https://github.com/motioneye-project/motioneye/blob/dev/motioneye/extra/linux_init) now, but actually I do not see much value for this, basically we should hardcode the `motion` version which is supported best by motionEye, which is `4.3` currently, since `4.4` has some config names changed. This also assures consistency across installs when `motioneye_init` is used, based on the motionEye version.

.github/workflows/docker.yml
MarcelCoding: Currently, the image is pushed to ghcr.io. Once the DockerHub stuff is solved we just have to add the corresponding tag to the meta action and remove the comment from the login to docker hub task. Therefore I see no problem to already start publishing to ghcr.io

MichaIng: Ah, I wasn't recognising. Found it: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-docker-registry

Okay, let's simply try it out. You are right that it helps with testing the current development as well, not only the Docker container in particular, when testers can pull a Docker container instead of needing to install motionEye natively. Then I suggest to use the `dev` branch, so it does not depend on us deciding to rename it, and we get a first image right when merging this PR.

Also these "packages" are shown on our GitHub page, below the releases, which is quite nice.

MarcelCoding:
Then I guess this is ready for review. If you don't have anything to add mark this pr as so. I am now going to sleep :sleeping:

MichaIng (approved):

MichaIng:
Hmm, the GitHub registry steps were skipped on `dev` branch merge. I'm not sure whether `(steps.meta.outputs.tags != null)` us valid syntax for checking whether it is defined, at least if not wrapped into `${{ ... }}`. If I'm not mistaken, the way it was simply having `&& steps.meta.outputs.tags` without `${{ ... }}` checks whether it is defined, at least I saw this in other workflows and docs.

MarcelCoding:
Maybe a check if the tags are just an empty string is also a possible solution?

MarcelCoding:
@MichaIng I just noticed that the ghcr.io image is private. https://github.com/motioneye-project/motioneye/pkgs/container/motioneye Maybe you could make it public?

MichaIng:
Many thanks for the hint. I changed it now to be public.

PR #1723, opened 2020-04-25
Added support for telegram motion notifications (UI & Backend).

HawtDogFlvrWtr:
Added support for telegram motion notifications. Added interface pieces for api, chat id, photo capture support, and testing. Instructions for setup are included as buttons and question mark info. Built on the current email implementation using all standard libraries or those already used within motionEye. Full UI integration.

PR #2402, opened 2022-04-05
Fix for #2264 saving Telegram settings

sirdel:
Fix for #2264 saving Telegram settings


MichaIng:
Many thanks for your contribution. Could you explain the rationale behind it?

It looks like the Telegram message line was added to `on_event_end` while it should have been added to `on_event_start`. It was of course meant as notification when motion happens, not when motion did not happen for a while (where we have the dedicated end event actions), so I'm wondering where this bug was introduced.

MichaIng:
Here it is, when the feature was added: #1723
It was moved to end events with a commit in the same PR, but further down the file it is looked for in `on_event_start` only, so looks like it was never functional? https://github.com/motioneye-project/motioneye/pull/1723/commits/0f80da6

I have no Telegram account (and do not plan to create one), so cannot test it myself.

@HawtDogFlvrWtr
Pinging you here to not break something intended:
- The code adds the Telegram event line to `on_event_end.append(line)`.
- But they are read back when looping through `for e in on_event_start:`.
- That doesn't seem to possibly work.
- You added commit https://github.com/motioneye-project/motioneye/pull/1723/commits/0f80da6 latest, so to me it looks like adjusting the other code properly has just been forgotten?
- But most importantly, was is the reason to have Telegram messages sent after motion stopped, instead of when motion starts? Other actions, like email, trigger on motion start, so it making this an end event action seems to break the logic, i.e. it not intuitive for users.

With #1914 we added a clearer separation between motion start and end event actions. So if it is seen worth, we could simply add two dedicated options for Telegram for start and end event.

motioneye/config.py
```diff
@@ -1148,6 +1148,20 @@ def motion_camera_ui_to_dict(ui, prev_config=None):
         )
 
         on_event_start.append(line)
+
```
MichaIng, line 1151:
```suggestion

```

sirdel:
> Many thanks for your contribution. Could you explain the rationale behind it?

Yep, the ui currently only caters for sending telegrams "on event start". The previous version "worked" but actually stored the telegram settings under "on event end" as a run command. This fix correct the miss-match between the ui and camera config file.

notifications

MichaIng (approved):

MichaIng:
Without getting further feedback, I'll merge this now. I don't see any reason why one would want Telegram messages to be sent after motion ended, instead of when motion started, and the first never worked as intended as was shown wrong in the UI. So with this change we have a reasonable and functional feature. An additional option to send a Telegram message after motion finished, can be added easily if required.

PR #2428, opened 2022-04-23
2355 make language selection a UI setting

jmichault:


motioneye/meyectl.py
```diff
@@ -24,33 +24,40 @@
...
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
-from motioneye import settings
+lingvo = 'eo'
+traduction = None
 
-_LOG_FILE = 'motioneye.log'
+# ŝarĝante tradukojn
+def load_l10n():
+    global lingvo, traduction
```
MichaIng, line 39: As of the CodeFactor/Pylint annotation: There is no way around global variables here, right? At least I can imagine it when those need to be enabled across all script for strings to be translated.

MichaIng:
Awesome, many thanks for implementing this. I cannot test it before next week, probably others can. See my comment about the CodeFactor/Pylint annotation. I can imagine that there is no way around global variables in this case, and to get CI green I can then mute this particular annotation.

motioneye/config.py
```diff
@@ -2142,6 +2149,7 @@ def _set_default_motion(data):
     data.setdefault('@admin_password', '')
     data.setdefault('@normal_username', 'user')
     data.setdefault('@normal_password', '')
+    data.setdefault('@lang', '')
```
MichaIng, line 2152: I suggest to use English as default language, understood by most users, and to not change the current behaviour:
```suggestion
    data.setdefault('@lang', 'en')
```

motioneye/handlers/config.py
```diff
@@ -214,6 +217,11 @@ def set_main_config(ui_config):
             reload = False
             restart = False
 
+            if lang != old_lang:
+                logging.debug('lang changed, restart needed')
+                # kill myself to force restart. Not very elegant...
+                os.kill(os.getpid(), 9)
```
MichaIng, line 223: Not very elegant as well, but indeed seems to be SIGKILL is the only signal where systemd triggers an automated restart. We may want to implement a specific signal or exit code and a related `Restart=` to the systemd unit.

motioneye/templates/main.html
```diff
@@ -161,6 +161,11 @@
                         <span class="minimize open"></span>
                     </div>
                     <table class="settings">
+                        <tr class="settings-item" strip="true">
+                            <td class="settings-item-label"><span class="settings-item-label">{{ _("Lingvo") }}</span></td>
+                            <td class="settings-item-value"><input type="text" autocapitalize="none" class="styled general main-config" id="langEntry"></td>
```
MichaIng, line 166: I suggest to make this a selection/list with all supported languages instead of an input box. So users do not need to guess which language is supported and how to enter it correctly (like `en` for English).

motioneye/templates/main.html
```diff
@@ -161,6 +161,11 @@
                         <span class="minimize open"></span>
                     </div>
                     <table class="settings">
+                        <tr class="settings-item" strip="true">
+                            <td class="settings-item-label"><span class="settings-item-label">{{ _("Lingvo") }}</span></td>
```
MichaIng, line 165: Let's add the translation for `Lingvo` directly. For German language it's `Sprache` 🙂.

motioneye/templates/main.html
```diff
@@ -161,6 +163,15 @@
...
+                            <td class="settings-item-value"><select class="styled general main-config" id="langSelect">
+                              {% for lang in settings.langlist %}
```
MichaIng, line 169: Ah the moment the list is empty. Not sure whether `settings` is available when the HTML is generated? Otherwise probably we just add a static list? Since language won't be added or removed often, it wouldn't be so hard to maintain.

jmichault: settings.langlist is populated in meyectl.load_l10n(), which is called before server.main(), so it's available when the HTML is generated.
Here is the result :
image

PR #2590, opened 2022-09-07
Fit vertically 2x2 as expected

### LINKED ISSUE #2583: Cameras previews don't fit to full browser window
With 4 cameras, I expected Layout Columns = 2, Layout Rows = 2 and Fit Frames Vertically enabled would make all previews scale to fit into equally sized quadrants as I resize the window. Is there anyway to achieve that behavior? I find myself having to scroll even when F11 my browser into full screen mode, because the bottom of the 2nd row doesn't fit the window.

lowlyocean:
Closes #2583

### DIFF
--- a/motioneye/static/css/main.css
+++ b/motioneye/static/css/main.css
@@ -917,6 +917,22 @@ div.page-container.four-columns div.camera-frame {
     width: 25%;
 }
 
+div.page-container.one-row div.camera-frame {
+    height: 100%
+}
+
+div.page-container.two-rows div.camera-frame {
+    height: 50%
+}
+
+div.page-container.three-rows div.camera-frame {
+    height: 33.3333%
+}
+
+div.page-container.four-rows div.camera-frame {
+    height: 25%
+}
+
 div.page-container div.camera-frame.full-screen-hidden {
     width: 0 !important;
     height: 0 !important;
@@ -1263,14 +1279,18 @@ div.camera-action-button.preset9:BEFORE {
 div.camera-container {
     position: relative;
     padding: 0;
+    height: 100%;
 }
 
 img.camera {
     position: relative;
-    width: 100%;
+    left: 50%;
+    top: 50%;
+    transform: translate(-50%, -50%);
+    height: inherit;
     display: block;
-    transition: opacity 0.2s ease;
-    opacity: 1;
+    transition: opacity .2s ease;
+    opacity: 1
 }
 
 img.camera.error,
--- a/motioneye/static/css/main.css
+++ b/motioneye/static/css/main.css
@@ -950,7 +950,7 @@ div.page-container div.camera-frame.single-cam-hidden {
 
 div.page-container div.camera-frame.single-cam {
     width: 100%;
-    height: auto;
+    height: inherit;
     transition: all 0.4s ease; /* prevents unwanted camera frame line wrapping */
 }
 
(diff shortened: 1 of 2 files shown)


MichaIng:
Many thanks. Please try to avoid trailing spaces, indentation changes and code style changes, to keep the code changes better reviewable. I'll do some tests when I find time.

zagrim (approved):
Looks sensible and likely working solution to me. The only things I'm somewhat wondering (and which could be quite easily tested, I admit) are:

1. How does it work when rows x cols is less than the number of configured cameras? Reshaping the camera array in that case seems to drop off the extra, will that happen to camera frames, too?
2. How does it work when there are more than four rows (or four columns for that matter, I see you took inspiration from the original column layout)? The number of "row" (and "column") CSS classes can be increased, but given the dynamic nature of the layout, should we instead have the height (and width) set with inline CSS so the code would work regardless of the setup size? Or is this a non-issue since the number of rows and cols is limited by the sliders in the UI? In that case it's all fine :smiley: 
4. How does it work when some cameras are in portrait orientation and some in landscape? Might not be very usual but some users have that kind of setup.

Please fix the indentations that were broken toward the end of main.js.

lowlyocean:
Thanks for reviewing. My editor complained that the tabs/spaces usage was inconsistent in the file, so I think the auto-fix just made it all spaces throughout. I can leave the unrelated lines of code the way they were

On #1 and #2, I don't think the UI sliders should even let you select rows x columns which isn't at least enough to accommodate the # of active cameras. And if you add/remove a camera, there should be a prompt that checks if your rows/columns choice needs to be re-adjusted accordingly. There isn't a sensible default behavior (you don't want to hide cameras, and you don't want to have extra empty rows/columns)

On #3, the more I think about this, more I think that the whole bit about reshaping and setting width / height of the container may not even be needed. What's doing the work here is the CSS that adjusts width to 33.3% if rows =3, 25% if rows = 4, etc. In that case orientation doesn't matter, as long as we take care that we check aspect ratio of each cam and fit to width if >1.0 else fit to height

lowlyocean:
Yeah, my editor wanted to make use of indents/spaces consistent throughout this file so it changed the unrelated lines. I can override that so it's just changing lines relevant to this issue

I agree, a mix of landscape/portrait cameras complicates things. Maybe let's tackle that later (right now, I think releasing this would be better than the current state of affairs even if not covering 100% of use cases - we can follow up if users complain). 

The UI probably shouldn't let you choose rows/columns that are inconsistent with the # of cameras chosen. Might have to fire an event when the user adds or deletes a camera, so that they adjust the rows/columns right then. There's no sensible default handling if you change # of cameras (you don't want to hide cameras, you don't want extra blank rows/columns with no cams)

zagrim:
Those points I raised were only things that popped up in my mind while reviewing the code, they might well be non-issues. At least what you wrote on them makes sense to me.

I agree that if by testing we can confirm that this PR generally improves things, even if leaving some earlier not-that-well handled edge cases dealt in a non-optimal way, it would make sense to merge. I don't even have any idea how to properly deal with great variations in the frame aspect ratios, other than maybe ordering them automatically so that similar frames are on the same row or column. But the users might anyway like to order the cameras themselves in their liking (and I think there's at least one issue for being able to change the order of the cams), so maybe there's not much use in trying to do anything about it here.

lowlyocean:
Pushed a new commit that reverts the unrelated whitespace change as discussed

Also noticed a regression relating to single-cam view not rendering correctly, so added a fix for it

PR #2698, opened 2023-02-03
Telegram optimizations

### LINKED ISSUE #2708: Telegram motion notifications not working
I have the following settings of my Motioneye

I'm trying to make motion notifications working but they refuse to. My Home Assistant is detecting motion from a curl command, so as i think motioneye is able to detect motion. 
If i click Test button below Telegram settings i recieve test message to my Telegram bot, but the are no notifications about motion detected and no still images. 
My motioneye is installed in docker container from this repo:

motionEye Version | 0.43.0
Motion Version | 4.5.1
OS Version | Linux 4.4.180+

Please help to make notifications work. Thanks for your support.

s17534:
Hello all, first of all, must say that motioneye is very advanced piece of software. Thanks to everyone involved to it.

As a user and beginning programmer wanted to share with everybody my code in which I changed the way TG notifications work. Let me say in a few words what's changed:

1. No longer missing pictures from notifications 
2. Ability to send in one batch many pictures regarding one specific motion detection, user can specify how much they want in a GUI (max 10 per event)
3. No need for searching picture files on disk, the list of them is gathered while current event is happening
4. Slight changes to motioneye's web components in regards to some descriptions, js etc
5. Changes are reflected in camera config, no need for user edit for TG to work 
6. Maybe more, but now I don't remember :)

Tested on two different motioneye's installs. Seems fine to me.

While working on the changes I found some bugs in motion and motioneye, but I probably should open issue as well.


s17534:
It seams that nobody is interested, :sadFace

MichaIng:
It is more that no one found time to really review and test the PR. What makes it difficult/raises efforts is that you did some major structural changes, so that the Telegram action now significantly differs form other actions, which may have implications that are hard to see now, but more importantly means that logic changes for all motionEye users, regardless whether Telegram is used or not. I think most of these changes are not needed for the actual purpose/issue you want to solve, and it would be much easier if you kept it hence minimal.

If there is a general problem the way how motion events/actions are handled right now, then we better think about generally changing this, instead of changing it for a single action and create inconsistency more difficulty for others to understand the general logic of the code.

Nuke142:
Want to test. Don't know how to. I'am using dev docker

s17534:
> I want to test. Don't know how to. I'am using dev docker. Telegram notifications a little broken. I use Attached Images Time=1; Capture mode=motion triggered (one); Motion Gap=1.
> 
> Image created when motion is started and telegram sends notification immediately, but without photo, because motion record still going. And this is a bug. If I set Attached Images Time=60 and record lasts few seconds - no problem, telegram shows photo too.
> 
> Thus, the notification comes without a photo if recording continues at the moment when the notification should come. Although the image is created in the first second, before the notification is sent.

Hi, it is a known limitation of current state of the send telegram part of the motioneye. That is why I made my own version. As you probably see in the above discussion, my pull request was denied and changes to the source code will not be merged into main branch. 

So in a such situation, the only (known to me) way to try out my approach to telegram notifications is to clone my repo and continue from there. I know there could be trust issues here, you may see for yourself what has been changed in the source and  if it is ok with you.

If after such deliberation you still want to try my proposal, feel free to drop me a message here. I will then assist you with said preparations.

MichaIng:
> As you probably see in the above discussion, my pull request was denied

We'll, I did not deny it, nor am I the only person in charge 😉. But I would indeed prefer the changes to be more minimal/targeted to the actual aim. This would make it also easier (faster) to properly review and test.

Logan-Fouts (approved):

Logan-Fouts (commented):
Worked on my instance but yes, Lots of changes happening all at once

zagrim (commented):
I tested this and took a quick peek at the code, too. It seems to work ok in the basic scenario at least (I didn't test any edge cases). 

There are some things to improve, which I tried to point out in the specific code parts. And there's one more, as this doesn't migrate or clean up the old Telegram notification config, and so the user is left with an extra "Run A Command" notification which does nothing. It's easy for the user to clean that up (and it kinda makes it easy to copy-paste the token and chat-id from there to the new one if the user is tech-savvy), but I think we should deal with that automatically within ME, since it shouldn't be too hard. That said, I don't know if there's any suitable commonly used mechanism for migrating web hooks so far...

(edit: ditch comment about splitting functionality to separate files since that is not really the case here)

What I can't really judge is the inclusion of `httpx`. If it is really superior to the old ways (pycurl, or urllib + plenty of wrapper code) like it looks to me on the quick look, it might be a good addition and something to go for. But not being a Python dev I can't judge.

Issue #2795, opened 2023-07-19
no previews for recorded videos on dev/edge docker


spupuz:
![image](https://github.com/motioneye-project/motioneye/assets/1721792/144fa394-b3b9-4a76-9b65-a6747d4f7fe7)

here is also attached the log

[_motioneye_logs.txt](https://github.com/motioneye-project/motioneye/files/12092802/_motioneye_logs.txt)

if anything else is needed i'm available

spupuz:
![image](https://github.com/motioneye-project/motioneye/assets/1721792/3433da13-337d-4525-8f7d-e8f62b657a5b)

zagrim:
Oh, I see that you've also reported this on the v0.43.0 release tracking issue (and also commented that reconfiguring cameras resolved the issue), but curiously there the problem was that "CUBIC" was not available, here it is now "LINEAR". But that detail isn't probably important.
It the docker image you are using motioneye-project/motioneye:edge ?

zagrim:
Reconfiguring the cameras didn't help here?

spupuz:
> Oh, I see that you've also reported this on the v0.43.0 release tracking issue (and also commented that reconfiguring cameras resolved the issue), but curiously there the problem was that "CUBIC" was not available, here it is now "LINEAR". But that detail isn't probably important. It the docker image you are using motioneye-project/motioneye:edge ?

yes it is



> Reconfiguring the cameras didn't help here?

all is working properly except preview of recorded image and also a friend of mine has the same error, the error i got before we i needed to reconfigure cam was related to stream not processed at all

zagrim:
Ok. The LINEAR vs CUBIC thing seems just to be about whether ME is trying to create thumbnail to be used in media picker, or get the "current" image for the UI camera frame (thus, streaming). But I find it curious if some reconfiguration of the camera got rid of the latter issue, since the error (in that other issue) was "module 'PIL.Image' has no attribute 'CUBIC'" and surely existence of Pillow attributes aren't dependent on camera configs.

I think that someone with expertise on debugging the Docker image build should take a look at this. Standalone dev branch build has no issue with either of the mentioned features, so it looks like there's a wrong "Pillow" being used. I don't have the time right now to start digging into this :disappointed:

Ryu945:
I don't remember the details but I tried changing file formats when I had issues with that.  I don't remember what I ended up fixing but I had less issues with things not being displayed.

zethis:
[...]

PR #2796, opened 2023-07-23
Fix no previews for recorded videos on dev/edge docker

zethis:
Fixes:  #2795

According to pillow documentation the constant Image.LINEAR was replaced with Image.BILINEAR and Image.CUBIC to  Image.BICUBIC

### DIFF
--- a/motioneye/mediafiles.py
+++ b/motioneye/mediafiles.py
@@ -909,7 +909,7 @@ def get_media_preview(camera_config, path, media_type, width, height):
     width = width and int(float(width)) or image.size[0]
     height = height and int(float(height)) or image.size[1]
 
-    image.thumbnail((width, height), Image.LINEAR)
+    image.thumbnail((width, height), Image.BILINEAR)
 
     bio = BytesIO()
     image.save(bio, format='JPEG')
@@ -1027,7 +1027,7 @@ def get_current_picture(camera_config, width, height):
     if width >= image.size[0] and height >= image.size[1]:
         return jpg  # no enlarging of the picture on the server side
 
-    image.thumbnail((width, height), Image.CUBIC)
+    image.thumbnail((width, height), Image.BICUBIC)
 
     bio = BytesIO()
     image.save(bio, format='JPEG')


zagrim:
The only thing I'm unsure with this is that whether we should go with what seems to be the intention of Pillow maintainers with the constants, and use Image.Resampling.BILINEAR and Image.Resampling.BICUBIC ([9.1.0 deprecations](https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#constants) refer to those, although 9.4.0 reverted deprecation of Image.BILINEAR and Image.BICUBIC along with some other top level constants).

zagrim:
In the effort of getting this merged before there is a ton of issues on the same topic, I requested @jmichault and @MichaIng for review. While this change is now nicely backwards-compatible (should work with <9.1. as well as >=10.0), do we want to have some version restrictions in requirements.txt ([see comment](https://github.com/motioneye-project/motioneye/issues/2795#issuecomment-1656590233))? If not, let's merge.

MichaIng:
The parameter is optional, default is `BICUBIC`, which has better downscaling quality but worse performance:
- https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.thumbnail
- https://pillow.readthedocs.io/en/stable/handbook/concepts.html#filters-comparison-table

According to the table `HAMMING` should be actually better for creating (downscaled) thumbnails, introduced with v3.4.0?

However, yeah, `BILINEAR` has not even a version info in the docs, so should be a valid replacement for all relevant Pillow versions. While `Image.Resampling.FILTER` is the correct new constants call since v9.1.0, for now there is no deprecation of the old `Image.FILTER`. However, to go future prove: Is there any possible issue to declare motionEye compatible with Pillow >=9.1.0 only? It has been released 1.5 years ago, and our dev install instructions + tests pull the newest version, so any other issue with it should have been recognised. Even for those rare systems where pre-compiled wheels are not available (32-bit ARM with Python versions not supported by piwheels), build dependencies are not intrusive: a C compiler and libjpeg and zlib headers.

spupuz:
@MichaIng  can you please merge on next docker image 0.43 dev please?

MichaIng:
Not sure what you mean? Docker images are generated on every `dev` commit automatically here: https://github.com/motioneye-project/motioneye/pkgs/container/motioneye

If you mean DockerHub, I think we'll not upload there until we did a stable 0.43 release, including a release of it on PyPI. At least that would be the logical order IMO, first stable code, then stable wheel, and stable container images last.

spupuz:
The 0.43 dev still has image preview broken.

Il Ven 15 Set 2023, 18:08 MichaIng ***@***.***> ha scritto:

zagrim:
> However, yeah, BILINEAR has not even a version info in the docs, so should be a valid replacement for all relevant Pillow versions. While Image.Resampling.FILTER is the correct new constants call since v9.1.0, for now there is no deprecation of the old Image.FILTER. However, to go future prove: Is there any possible issue to declare motionEye compatible with Pillow >=9.1.0 only? It has been released 1.5 years ago, and our dev install instructions + tests pull the newest version, so any other issue with it should have been recognised. Even for those rare systems where pre-compiled wheels are not available (32-bit ARM with Python versions not supported by piwheels), build dependencies are not intrusive: a C compiler and libjpeg and zlib headers.

I was going toward being as backward-compatible as easily possible since I don't think we know really how much ME is being used on old RPi installations etc. Then again, that would expose us to having to deal with this same thing again when Pillow team decides to deprecate (again) `Image.BILINEAR` in favour of `Image.Resampling.BILINEAR`. Not being a Python developer I don't know if fixing the version to >=9.1.0 would only require some extra dependencies and the build would then be handled by pip automatically or if we'd end up writing extra instructions for how to build Pillow on those supposedly rare systems that don't have recent enough version.

OTOH since this version compatibility thing is really not an issue that needs to be tackled at this moment, I was already considering just merging this PR in the lack of opinions from the other maintainers, and leaving all this other stuff for the later time since there are folks now who have broken thumbnailing functionality.

MichaIng:
> the build would then be handled by pip automatically

The compiler and build dependencies would need to be installed manually, if running a host system where PyPI and piwheels do not provide pre-compiled wheels (which `pip` pulls automatically otherwise). But for now I agree that we should at least have some more system on v0.43 (resp. more away from 0.42) via stable release first, and can think about dropping support for old modules at a later time. The change is a small one, so I think I won't forget. I just watched the Pillow repo to be informed about releases, so I can recognise immediately if this breaking change is finally done.

> leaving all this other stuff for the later time since there are folks now who have broken thumbnailing functionality.

👍

zethis:
Do you want me to to use  `Image.Resampling.BILINEAR` instead of `Image.BILINEAR` ?

MichaIng:
Let's leave it for now.

zethis:
just tested locally, fix the issue too
image

zagrim:
Let's get this merged finally to get those previews fixed, then deal with Pillow version compatibility separately.

zagrim (approved):

Issue #844, opened 2018-03-28
number of action buttons limited?


hobbyquaker:
I can not add more than 7 action buttons, only 8 including the "take snapshot" button are shown. Is that correct? If yes - is there a config option to show more?

PS. Great project, thank you so much for your work! :)

ccrisan:
The place there is limited to 8 slots. You can't have more than 8 buttons at once.

PR #2903, opened 2023-12-28
Increase amount of action buttons 

sevimuelli:
Hi

I made some UI changes to increase the available action buttons. There have been several requests concerning this. This pull request fixes: #613 ,#844, #1766. 

Screenshot 2023-12-28 at 23 55 39
Screenshot 2023-12-28 at 23 57 22
Screenshot 2023-12-28 at 23 57 52
Screenshot 2023-12-28 at 23 58 19

I moved the frame rate and monitoring info to the top. I tried to optimize everything for space constraints. There are two rows in the top left. First one for camera name and frame rate and the second one for monitoring info. If there is no monitoring info, the second row disappears and name and fps are centered. The bottom is now solely for action buttons. Instead of 8, it allows up to 16. If there are only a few or if the window is wide enough, there is only one row. If more are added or the view changes and the camera window gets smaller, a second row appears. 

There are no breaking changes or anything, simply a UI enhancement. The only thing I am not super happy with for now is that in order to be able to make the buttons evenly spaced ( I used flex), I need to check the width of the camera window and make the settings in css accordingly. To do so, I check on a regular basis for the width. The only place I have found so far is in the main.js where I found a function that get's executed once a second to update frame rate and stuff. This is the main code for on line 5096 in main.js:

```
    /* check if action buttons need one or two rows*/
    if(cameraConfig.actions.length) {
        var actionButtonWidth = actionButtonDict[cameraConfig.actions[0]].width();
        var actionsWidth = actionsCount * actionButtonWidth;
        var cameraWidth = cameraOverlay.find('div.camera-overlay-bottom').width();
        var flexBasis = 200 / actionsCount;
        flexBasis += '%';
        if(cameraWidth <= actionsWidth){
            actionButtonWrapper.css('flex', '1 0' + flexBasis);
        } else {
            actionButtonWrapper.css('flex', 'none');
        }
    }
```
But I don't need to check it regularly, only when the UI changes. There is this function "updateLayout();" in main.js, which gets called only when the view changes but I don't see a way to access the individual data of each camera-window inside this function. Any recommendations are welcome. But I think for now, this isn't to big of a problem.


MichaIng:
Many thanks, this looks reasonable.

Could you show some screenshots form before your changes? I was actually wondering how to enable all those buttons. I can add one to create still images (when enabled), but the others are, I guess, added depending on the camera device, whether it supports zooming, moving, a light, or other functions?

Previously only the camera name and motionEye buttons are at the top, the framerate at the button left and the action buttons at the bottom right. It makes sense that they can exceed the camera frame, or do some ugly line break when being too many.

I wonder whether it makes sense to leave the camera name larger in one row, and the framerate and optional monitoring info in two rows, like it was before? For identifying reasons, the camera name might better large. Or alternatively, probably the best solution, have the framerate below the camera name, as it is always available, and the monitoring info as optional second column. That way also on very small camera frames, if no monitoring info is shown, there is more space for all info at the top. It is a different topic, but in my case, and I see the same in one of your cases, the monitoring info is an empty string, and shown as `""`, instead of just hidden. We should just set it empty, to have it omitted completely, in this case.

_Side note: The CodeFactor annotations are unrelated to this PR and, AFAIK, wrong, since those functions/variables are in fact used from HTML elements._

motioneye/static/css/main.css
```diff
@@ -1026,24 +1027,74 @@ div.mask-element.on {
...
     vertical-align: top;
+    text-align: left;
+    line-height: 1em;
+    flex-grow: 2;
+    flex-direction: column;
+}
+
+div.camera-info-top {
+    display: flex;
+    flex-grow: 2;
```
MichaIng, line 1046: Remove trailing space:
```suggestion
    flex-grow: 2;
```

motioneye/static/js/main.js
```diff
@@ -5084,6 +5093,20 @@ function addCameraFrameUi(cameraConfig) {
...
+        var cameraWidth = cameraOverlay.find('div.camera-overlay-bottom').width();
+        var flexBasis = 200 / actionsCount;
+        flexBasis += '%';
+        if(cameraWidth <= actionsWidth){
+            actionButtonWrapper.css('flex', '1 0' + flexBasis);
+        } else {
+            actionButtonWrapper.css('flex', 'none');
+        }
+    }
+
```
MichaIng, line 5109: Remove trailing space:
```suggestion

```

motioneye/static/js/main.js
```diff
@@ -4973,13 +4980,15 @@ function addCameraFrameUi(cameraConfig) {
...
-    }
+    var actionsCount = Math.ceil(cameraConfig.actions.length);
+
+    // if (cameraConfig.actions.length <= 4) {
+    //     cameraOverlay.find('div.camera-overlay-bottom').addClass('few-buttons');
+    // }
+    // else {
+    //     cameraOverlay.find('div.camera-action-buttons-wrapper').css('width', Math.ceil(cameraConfig.actions.length / 2) * 2.5 + 'em');
+    //     //cameraOverlay.find('div.camera-action-buttons-wrapper').css('width', 4 * 2.5 + 'em');
+    // }
```
MichaIng, line 4992: Let's do not keep old code commented. No one will ever be able to understand how this was used, with the CSS changed. And if someone wants to know how it worked before, on Git(Hub) the PR/commits can be checked, or an older commit checked out:
```suggestion
```

motioneye/static/js/main.js
```diff
@@ -4735,41 +4745,41 @@ function addCameraFrameUi(cameraConfig) {
...
-                                '<div class="button icon camera-action-button mouse-effect preset preset3" title="preset 3"></div>' +
-                                '<div class="button icon camera-action-button mouse-effect preset preset4" title="preset 4"></div>' +
-                                '<div class="button icon camera-action-button mouse-effect preset preset5" title="preset 5"></div>' +
-                                '<div class="button icon camera-action-button mouse-effect preset preset6" title="preset 6"></div>' +
-                                '<div class="button icon camera-action-button mouse-effect preset preset7" title="preset 7"></div>' +
-                                '<div class="button icon camera-action-button mouse-effect preset preset8" title="preset 8"></div>' +
-                                '<div class="button icon camera-action-button mouse-effect preset preset9" title="preset 9"></div>' +
+                                '<div class="camera-action-button-wrapper"><div class="button icon camera-action-button mouse-effect lock" title="lock"></div></div>' +
```
MichaIng, line 4750: I think it should be possible to skip the `camera-action-button-wrapper` div, isn't it? Or what is the purpose of centering a single div within its own parent div?

motioneye/static/js/main.js
```diff
@@ -5084,6 +5093,20 @@ function addCameraFrameUi(cameraConfig) {
...
 
+    /* check if action buttons need one or two rows*/
+    if(cameraConfig.actions.length) {
+        var actionButtonWidth = actionButtonDict[cameraConfig.actions[0]].width();
+        var actionsWidth = actionsCount * actionButtonWidth;
+        var cameraWidth = cameraOverlay.find('div.camera-overlay-bottom').width();
+        var flexBasis = 200 / actionsCount;
+        flexBasis += '%';
+        if(cameraWidth <= actionsWidth){
+            actionButtonWrapper.css('flex', '1 0' + flexBasis);
```
MichaIng, line 5104: Ah wait, now I do get an error in browser console about this:
```
main.js?v=0.43.1b1:5104 Uncaught ReferenceError: actionButtonWrapper is not defined
    at addCameraFrameUi (main.js?v=0.43.1b1:5104:13)
    at updateCameras (main.js?v=0.43.1b1:5133:13)
    at recreateCameraFrames (main.js?v=0.43.1b1:5148:9)
    at HTMLImageElement.<anonymous> (main.js?v=0.43.1b1:3396:21)
    at r.complete (jquery.min.js?v=0.43.1b1:2:62910)
    at c (jquery.min.js?v=0.43.1b1:2:28327)
    at Object.fireWith [as resolveWith] (jquery.min.js?v=0.43.1b1:2:29072)
    at u (jquery.min.js?v=0.43.1b1:2:59642)
    at S.fx.tick (jquery.min.js?v=0.43.1b1:2:64608)
    at ot (jquery.min.js?v=0.43.1b1:2:58967)
```
And here the asset versioning is visible, so all good in this regards 🙂.

motioneye/static/js/main.js
sevimuelli: Yes off course. Sry about that. Done

sevimuelli:
This is how it looked before:
localhost_8081_
To enable the buttons one simply needs to add the corresponding files under /etc/motioneye. For example up_1, down_1, preset1_1 and so forth. For testing, they don't even need to do something. An empty file that is executable is enough to show up in motioneye. 

I think your idea with the optional second column is the best idea. I will implement that. I noticed that on small screens, it is very crowded. i fixed the issue with the monitoring info. The problem was that even if there is no info, the string contains "". So the value was "\"\"". Now this case will be ignored. 

Considering the CodeFactor, this is actually correct. The four values showEmbedUrl, authorizeUpload, showMjpgUrl, showSnapshotUrl can't be found because as you said, they are used in html. There I added the comment "// eslint-disable-next-line no-unused-vars" so they don't throw an error anymore. But the other two, layoutRows and remCameraFrameUi are indeed not used. They were used before but the code changed and it was not removed. I commented those out with a remark and will clean it up properly in a next pull request.

motioneye/static/js/main.js
sevimuelli: The reason why I need a wrapper for every single button is that I wanted to space all buttons evenly. So each element needs to be able to grow. Without this, it would look like this:

And I can't do this on the underlying div that contains the action button itself because they contain a background-image of the buttons. So when you change the dimension of this div, it would look something like this:

sevimuelli:
I moved the monitoring info to a second column which only appears if needed. 
Screenshot 2023-12-31 at 13 16 38
Screenshot 2023-12-31 at 13 17 03
Screenshot 2023-12-31 at 13 16 50


Is this how you imagined it or would you like me to change something?

motioneye/static/js/main.js
MichaIng: Understood. So in case we want to reduce the amount of nested elements, we'd need to do more changes to how the buttons are presented, like splitting the images at best into individual svg paths, directly inline attached to an individual class each. Much cleaner anyway than this strange background positioning. But something we can address in a dedicated PR when someone finds time and mood for some non-visual optimisations.

But more importantly, what about the missing `actionButtonWrapper`? That one still needs to be defined somewhere, or CSS selector used directly like in my suggestion.

MichaIng:
> I moved the monitoring info to a second column which only appears if needed.

I indeed meant a "column", not a "row" what you added now. So basically switching framerate and monitoring info positions form how the PR was at first, and then have this 2nd column removed if no monitoring info is available.

But actually I like the solution with a 3rd row more, respectively a 2nd row considering how elements are nested. Allow text information listed below each other IMO is clearer, and I think this little additionally removed/overlaid camera view is worth it. So from my end: Let's keep what you have now 👍.

motioneye/static/js/main.js
sevimuelli: Oh, don't know why but it got deleted for some reason. I put it back in. 

So I played with CSS again a bit but I see no way that one can do it without a wrapper as long as all this icons are in one image. It would be easy if the buttons were a fixed size all the time but because the width of the wrapper changes depending on the button count, it is not possible like this. Right now every button has it's own class already. I think it wouldn't be too big of a trouble to fix it. I can do it in the next PR if you want.

sevimuelli: So I ended up splitting the button SVG after all, so we have a clean solution now instead of one more thing to think about. So I could also remove the `actionButtonWrapper `. All in all a lot cleaner.

sevimuelli:
So from my standpoint I think everything is completed so far. If you have any other ideas / improvements I'm happy to implement it. 

Concerning you points about the CodeFactor issues, thank you for clarifying this. I would like to do this properly to check for any leftovers, not just delete the blocks I commented out. As soon as this PR is merged, I will do a proper cleanup and open a new PR. I really want to check where all those variables are used and what can be removed without removing too much. 😅

motioneye/static/js/main.js
```diff
@@ -29,7 +29,8 @@ var pageContainer = null;
 var overlayVisible = false;
 var layoutColumns = 1;
 var fitFramesVertically = false;
-var layoutRows = 1;
+// Was removed 5 years ago, needs cleanup, a25dec205b70fbf0f75f8c2f424d4920a875f399
+// var layoutRows = 1;
```
MichaIng, line 33: Let's clean this up already:
```suggestion
```

motioneye/static/js/main.js
```diff
@@ -5084,17 +5088,47 @@ function addCameraFrameUi(cameraConfig) {
...
-}
+// was removed 10 years ago, needs cleanup, a862509fa816049d1fa6d7da1712b2a6c5e34519
+// function remCameraFrameUi(cameraId) {
+//     var cameraFrameDiv = getPageContainer().find('div.camera-frame#camera' + cameraId);
+//     cameraFrameDiv.animate({'opacity': 0}, 100, function () {
+//         cameraFrameDiv.remove();
+//     });
+// }
 
 function recreateCameraFrames(cameras) {
```
MichaIng, line 5133:
```suggestion
function recreateCameraFrames(cameras) {
```

MichaIng:
Great work!

Let's do the cleanup right away. We have this PR as reference and the commits you referenced for the obsolete variables/functions remain visible here.

I'll do a final test this weekend and merge it if everything works well.

motioneye/static/js/main.js
sevimuelli: The thing here is that the variable is not really used because it's function was removed but there are still some functions around where the variable is set:
```
$('#layoutRowsSlider').change(function () {
        layoutRows = parseInt(this.value);
        updateLayout();
        savePrefs();
    });
```
and

```
function applyPrefs(dict) {
    setLayoutColumns(dict['layout_columns']);
    fitFramesVertically = dict['fit_frames_vertically'];
    layoutRows = dict['layout_rows'];
    framerateFactor = dict['framerate_factor'];
    resolutionFactor = dict['resolution_factor'];

    if (fitFramesVertically) {
        getPageContainer().addClass('fit-frames-vertically');
    }
    else {
        getPageContainer().removeClass('fit-frames-vertically');
    }
}
``` 
I think those are some leftovers from the original functionality. It looks like there are a couple of things around from this one. Wouldn't it make sense to remove all those things that are not longer used anymore instead of deleting only this variable? I can do that tomorrow if you'd like?

motioneye/static/js/main.js
sevimuelli: Ok this is really not used anymore. I removed it 👍

motioneye/static/js/main.js
MichaIng: Ah right, I forgot about that. There is even the GUI toggle to set this, which should be removed all together. But we can do that in a dedicated PR.

sevimuelli:
I moved the calculation for the action buttons to a separate callback function which only gets executed when the layout updates. Before it was executed once a second which is quite unnecessary. But now really everything is done, so you can test it. Thank you.

sevimuelli:
Hi, could you do the testing and merge this weekend, so I could to the cleanup in the next step as we talked about?

MichaIng (approved):
Looks and works all as expected. Many thanks!

PR #3192, opened 2025-08-31
Support S3 Server Side Encryption with custom key

tuxgasy:
Add a supports of Server Side Encryption known as [SSE-C](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html) for compatible S3 bucket provider.

The encryption is performed on the S3 server with the provided key. The key is not stored on the S3 server. To download encrypted file, you must therefore provide the same key.

The encryption key must be a base64 encoded string with a length of 32 bytes. You can generate one as following:

`openssl rand 32 | base64`

### DIFF
--- a/motioneye/uploadservices.py
+++ b/motioneye/uploadservices.py
@@ -1211,7 +1211,7 @@ class S3(UploadService):
             self._secret_key = data['secret_key']
         if data.get('bucket') is not None:
             self._bucket = data['bucket']
-        if data.get('sse_c_key'):
+        if data.get('sse_c_key') is not None:
             self._sse_c_key = data['sse_c_key']
 
     def upload_file(self, target_dir, filename, camera_name):
--- a/motioneye/uploadservices.py
+++ b/motioneye/uploadservices.py
@@ -23,7 +23,8 @@ import mimetypes
 import os
 import os.path
 import time
-from base64 import b64encode
+from base64 import b64encode, b64decode
+from hashlib import md5
 from urllib.error import HTTPError
 from urllib.parse import quote, urlencode
 from urllib.request import Request
@@ -1189,6 +1190,7 @@ class S3(UploadService):
         self._access_key = None
         self._secret_key = None
         self._bucket = None
+        self._sse_c_key = None
         UploadService.__init__(self, camera_id)
 
     def dump(self):
@@ -1197,6 +1199,7 @@ class S3(UploadService):
             'access_key': self._access_key,
             'secret_key': self._secret_key,
             'bucket': self._bucket,
+            'sse_c_key': self._sse_c_key,
         }
 
     def load(self, data):
@@ -1208,6 +1211,8 @@ class S3(UploadService):
             self._secret_key = data['secret_key']
         if data.get('bucket') is not None:
             self._bucket = data['bucket']
+        if data.get('sse_c_key'):
+            self._sse_c_key = data['sse_c_key']
 
     def upload_file(self, target_dir, filename, camera_name):
         # Create an S3 client
@@ -1226,12 +1231,22 @@ class S3(UploadService):
         else:
             rel_filename = os.path.basename(filename)
 
+        if self._sse_c_key:
+            sse_key_md5 = b64encode(md5(b64decode(self._sse_c_key)).digest()).decode("utf-8")
+            extra_args = {
+                "SSECustomerAlgorithm": "AES256",
+                "SSECustomerKey": self._sse_c_key,
+                "SSECustomerKeyMD5": sse_key_md5,
+            }
+        else:
+            extra_args = None
+
         # Uploads the given file using a managed uploader, which will split up
         # large files automatically and upload parts in parallel.
         self.debug(
             f'uploading file "{filename}" to S3 bucket "{self._bucket}" path "{rel_filename}"'
         )
-        s3.upload_file(filename, self._bucket, rel_filename)
+        s3.upload_file(filename, self._bucket, rel_filename, ExtraArgs=extra_args)
 
     def test_access(self):
         try:
--- a/motioneye/uploadservices.py
+++ b/motioneye/uploadservices.py
@@ -23,7 +23,7 @@ import mimetypes
 import os
 import os.path
 import time
-from base64 import b64encode, b64decode
+from base64 import b64decode, b64encode
 from hashlib import md5
 from urllib.error import HTTPError
 from urllib.parse import quote, urlencode
@@ -1232,11 +1232,13 @@ class S3(UploadService):
             rel_filename = os.path.basename(filename)
 
         if self._sse_c_key:
-            sse_key_md5 = b64encode(md5(b64decode(self._sse_c_key)).digest()).decode("utf-8")
+            sse_key_md5 = b64encode(
+                md5(b64decode(self._sse_c_key)).digest()  # nosec B324
+            ).decode()
             extra_args = {
-                "SSECustomerAlgorithm": "AES256",
-                "SSECustomerKey": self._sse_c_key,
-                "SSECustomerKeyMD5": sse_key_md5,
+                'SSECustomerAlgorithm': 'AES256',
+                'SSECustomerKey': self._sse_c_key,
+                'SSECustomerKeyMD5': sse_key_md5,
             }
         else:
             extra_args = None
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -513,6 +513,7 @@ def get_camera(camera_id, as_lines=False):
             '@upload_access_key',
             '@upload_secret_key',
             '@upload_bucket',
+            '@upload_sse_c_key',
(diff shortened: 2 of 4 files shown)


tuxgasy:
MD5 hash of the encryption key is required. How to ignore the failing check ?

motioneye/uploadservices.py
```diff
@@ -1221,10 +1226,20 @@ def upload_file(self, target_dir, filename, camera_name):
             aws_secret_access_key=self._secret_key,
         )
 
+        if self._sse_c_key:
+            sse_key_md5 = b64encode(md5(b64decode(self._sse_c_key)).digest()).decode("utf-8")
+            extra_args = {
+                "SSECustomerAlgorithm": "AES256",
+                "SSECustomerKey": self._sse_c_key,
+                "SSECustomerKeyMD5": sse_key_md5,
```
MichaIng: Is it possible to pass the key with a stronger hashing algorithm?

MichaIng: Oh it actually seems this is not needed but calculated automatically if missing: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/multipartupload/complete.html
> Please note that this parameter is automatically populated if it is not provided. Including this parameter is not required

motioneye/uploadservices.py
```diff
@@ -1221,10 +1225,23 @@ def upload_file(self, target_dir, filename, camera_name):
...
+            extra_args = None
+
         # Uploads the given file using a managed uploader, which will split up
         # large files automatically and upload parts in parallel.
         self.debug(f'uploading file "{filename}" to S3 bucket "{self._bucket}"')
-        s3.upload_file(filename, self._bucket, filename[len(target_dir) :])
+        s3.upload_file(
+            filename,
+            self._bucket,
+            filename[(len(target_dir)+1) :],
```
MichaIng: The `+1` was added to remove the last slash as well? Did this always appear in the resulting key? But what happens if the `target_dir` is passed with trailing slash? I am currently not sure whether it is striped somewhere, otherwise we could either `target_dir.rstrip('/')` or alternatively `filename[len(target_dir) :].lstrip('/')` to achieve the same effect.

MichaIng:
@tuxgasy 
Can you please test it with the `SSECustomerKeyMD5` argument removed? As said, according to boto3 docs, it is not needed, but calculated automatically if missing. And see my above question regarding `+1` removed characters for the key name.

motioneye/uploadservices.py
MichaIng: Ohh, it seems to be actually solved better here: https://github.com/motioneye-project/motioneye/pull/2760/changes
Let me merge this first.

tuxgasy:
> @tuxgasy Can you please test it with the `SSECustomerKeyMD5` argument removed? As said, according to boto3 docs, it is not needed, but calculated automatically if missing. And see my above question regarding `+1` removed characters for the key name.

Without SSECustomerKeyMD5, I get the following error message :

```
boto3.exceptions.S3UploadFailedError: Failed to upload /var/lib/motioneye/Camera1/2026-02-04/18-07-57.jpg to me-salon/2026-02-04/18-07-57.jpg:
An error occurred (InvalidArgument) when calling the PutObject operation: The secret key was invalid for the specified algorithm.
```

MichaIng:
This is latest boto3? Weird, the docs say it is not needed: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/upload_part.html
Though this is `upload_part` and a bunch of other methods: https://boto3.amazonaws.com/v1/documentation/api/latest/search.html?q=SSECustomerKeyMD5
But the `upload_file` page does not explicitly mention it: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/upload_file.html

However, if it does not work reliably without, I'll add it back in.

Since I want to fix/merge #2760 first, can you confirm that using slashes in the filename creates subdirectories on the S3 server?

tuxgasy:
> Since I want to fix/merge #2760 first, can you confirm that using slashes in the filename creates subdirectories on the S3 server?

The subdirectories are created on S3 server. But when I disable `Include Subdirectories`, I get the following error :

```
motioneye-1  |    ERROR: failed to upload file "/var/lib/motioneye/Camera1/2026-02-05/06-46-43.jpg" with service s3: object of type 'bool' has no len()
motioneye-1  | Traceback (most recent call last):
motioneye-1  |   File "/usr/local/lib/python3.13/dist-packages/motioneye/uploadservices.py", line 1324, in upload_media_file
motioneye-1  |     service.upload_file(target_dir, filename, camera_name)
motioneye-1  |     ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
motioneye-1  |   File "/usr/local/lib/python3.13/dist-packages/motioneye/uploadservices.py", line 1242, in upload_file
motioneye-1  |     filename[(len(target_dir) + 1) :],
motioneye-1  |               ~~~^^^^^^^^^^^^
motioneye-1  | TypeError: object of type 'bool' has no len()
```

We need to check if `target_dir` is a boolean (False ?). Do you want to fix it with the other PR ?

MichaIng:
Yes, fixing this was the purpose of the other PR. Fixing the possible leading slash (as well with preserved subdirs enabled) was a side effect. The PR has been merged and this one rebased.

I restored the MD5 sum (and found the correct way to let `bandit` ignore the `md5()` call as weak algorithm 😅). Can you do a final test please?

MichaIng:
It works for me with a MinIO test server. Since SSE-C requires HTTPS, and I generated a self-signed HTTPS certificate, I needed to tell boto3 to not verify the certificate:
```py
            s3 = boto3.client(
                's3',
                endpoint_url=self._endpoint_url,
                aws_access_key_id=self._access_key,
                aws_secret_access_key=self._secret_key,
                verify=False,
            )
```
Might be an idea to add this as another toggle in motionEye. Although it does not seem to be practical at all, since most clients will deny access by default if the certificate cannot be verified. Leaving this as test/dev-only case.

Now the uploaded files are a tiny bit larger, and I cannot download or "preview" them anymore with the MinIO community server web UI 😅:

image

However, I was able to download them with the MinIO command-line client:
```
mc get --debug --insecure --enc-c "local/test/2026-02-09/19-21-12.jpg=Qa/Tz+fWqCZU7yrlENFM469j0/8WzDr3WHWVAtlBPY8" local/test/2026-02-09/19-21-12.jpg 19-21-12.jpg
```
_The trailing `=` needs to be removed from the base64 key for `mc` to accept it. Using any other key leads to a 403 response, which explains why the web UI cannot access them. I do not see a way to configure encryption keys there, it has been crippled since MinIO wants to push free users to their pair product._

And I can replicate as well that the MD5 is strictly needed. Weird that the boto3 docs mention it wouldn't be needed for all low-level commands, while the high-level command `upload_file()` obviously requires it.

@tuxgasy
What I am wondering now:
- SSE-C does not require any configuration at the server side: the client passes the encryption key, and any client requires the same key to download those files: https://minio.community/community/minio-object-store/administration/server-side-encryption/server-side-encryption-sse-c.html
- Do you think it is obvious enough for users that the generated key must be stored elsewhere and used to access the uploaded media? Otherwise, if they every remove the key from motionEye, all uploaded motionEye media on the S3 is practically lost. Otherwise we might want to make that clear in the tooltip.
- And we might either want to add an `(optional)` to the setting's name, or hide it behind a toggle: `use_sse_c` and `sse_c_key` as dedicated settings. Just so that it is clear that this field needs to be filled only if SSE-C is actually wanted. Alternatively, wasn't there a way to show some greyed out text in HTML forms? Not actually pre-filling them, just as informational text, so users see directly that it is optional, without hovering the question mark first. ... found it, the `placeholder` attribute:
  ```html
  placeholder="{{ _("optional base64-encoded SSE-C encryption key") }}"
  ```

MichaIng (approved):
Looks and works good for me now.

MichaIng:
Btw, AWS is trying to drop SSE-C support 😄: https://aws.amazon.com/blogs/storage/advanced-notice-amazon-s3-to-disable-the-use-of-sse-c-encryption-by-default-for-all-new-buckets-and-select-existing-buckets-in-april-2026/

Anyway, for now, it can be still re-enabled if needed. What I like about it is that it does not require and server-side configuration and no dedicated KMS server. That way, it is simple. But from user perspective, it works like client-side encryption, hence I wonder how complicated it would be to use that instead. There are SDKs, but no Python library: https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingClientSideEncryption.html
An ancient boto3 wrapper: https://github.com/boldfield/s3-encryption
I mean any encryption tools can be used, but doing it a way that it is compatible with the AWS SDK and client is the challenge.

Anyway, something for a dedicated PR if anyone finds time to dig deeper into this topic.

PR #3252, opened 2026-01-03
Create python-package-conda.yml

lyhbee:


MichaIng:
Is this some LLM commit? Linting is done via `pre-commit`, and `pytest` runs already in other workflows. So these two are unnecessary steps. If you want to somehow test a motionEye installation via `conda`, then it looks like that step is missing. But I am not sure whether this has any point: If it installs fine via `pip`, it should install fine with any other Python package manager, i.e. mostly a waste of data center energy to test all of them separately.

PR #3253, opened 2026-01-06
Fix current picture retry logic

### LINKED ISSUE #2978: 0byte snapshot files - no new releases and bugfixes?
Version 0.43.1b1 creates 0byte files for some cameras with the snapshot request URL. This bug is known. It seems as if motioneye is unfortunately no longer being developed further.

### LINKED ISSUE #2989: snapshot url friert ein
Hallo, leider friert bei der installation Methode die Snapshot url ein wenn man die Browser Seite von MotionEye schließt, ich benutze sie für mein enigma2 receiver

### LINKED ISSUE #3054: wget of still images not working the first time
For the last 3 years, I've been running MotioneyeOS (from June 6, 2020) in a Raspberry Pi Zero W with a RPi camera module attached via the CSI camera connector.
motionEye Version	0.42.1
Motion Version	4.2.2+gitUNKNOWN
OS Version	motionEyeOS 20200606
I've been using a simple bash script run by a CRON job entry to capture images, for my weewx weather station page.  
This is the line in the script that does the capture:
`wget http://192.168.7.90/picture/1/current/ -O /home/weewx/public_html/images/webcam.jpg`
It only calls for an image 1 time.
Doing this has worked great, once per minute, for 3 years.
(Note, that with MotioneyeOS you don't need the port :8765 notation in your wget call.)
This week, I loaded some other Raspberry Pi Zero boards (with attached RPi camera modules) loaded with 32-bit Bullseye RaspianOS from July 2024.  
motionEye Version	0.43.1b2
Motion Version	4.7.0
OS Version	Raspbian 11
I got Motioneye working in Bullseye and thought I had working Raspberry Pi cameras that were working as well as my 2020 vintage MotioneyeOS RaspberryPi cameras.  Nope!
I found that the same wget command line I was using with the MotioneyeOS RPi cameras, does NOT work the same with the current Motioneye release running inside Bullseye RaspbianOS.  
I found that the wget command doesn't grab a valid image from Motioneye, UNLESS you call wget, wait 2 seconds, and then call wget AGAIN!
It's like Motioneye puts the camera to sleep and it takes 2 wget calls to wake it up and get a valid image??
By experimenting, I found that calling wget 2 times with a 2 second delay between calls, wakes up the camera and obtains a valid image.
If I don't do this, I get a 0 byte .jpg file.
```
wget http://192.168.7.91:8765/picture/1/current/ -O /dev/null
sleep 2
wget http://192.168.7.91:8765/picture/1/current/ -O /home/weewx/public_html/images/webcam.jpg
```
Why is it now necessary to call wget 2 times to get a valid image from the camera with the latest release of Motioneye?
Is the current Motioneye putting the camera module to sleep and thus it is necessary to issue 2 wget calls to wake it up?
If so, is there a setting that will keep Motioneye from putting the camera to sleep?

Help appreciated,

Marijn0:
Fixes an issue where /picture/\<id\>/current did not return JPEG images when no frame was available yet.

The retry logic previously used `IOLoop.add_timeout()`, which scheduled a retry but immediately closed the HTTP request.
As a result, the client could receive a response with Content-Type: image/jpeg but an empty body.

The retry now runs within the same request using an awaited sleep, keeping the request open while waiting for the first JPEG frame from the internal MJPEG client, up to the intended retry limit.

This behavior matches the existing comment in the code, and I think it reflects the intended behavior?:

https://github.com/motioneye-project/motioneye/blob/4f1b7b50182a6a23cddf3fa88fb0b40f0e667a20/motioneye/handlers/picture.py#L114-L116

#2978

### DIFF
--- a/motioneye/handlers/picture.py
+++ b/motioneye/handlers/picture.py
@@ -20,7 +20,7 @@ import logging
 import os
 import re
 
-from tornado.ioloop import IOLoop
+from tornado import gen
 from tornado.web import HTTPError
 
 from motioneye import (
@@ -115,12 +115,8 @@ class PictureHandler(BaseHandler):
             # get_current_picture() will make sure to start a client, but a jpeg frame is not available right away;
             # wait at most 5 seconds and retry every 200 ms.
             if not picture and retry < 25:
-                return IOLoop.current().add_timeout(
-                    datetime.timedelta(seconds=0.2),
-                    self.current,
-                    camera_id=camera_id,
-                    retry=retry + 1,
-                )
+                await gen.sleep(0.2)
+                return await self.current(camera_id=camera_id, retry=retry + 1)
 
             self.set_cookie(
                 'motion_detected_' + camera_id_str,


MichaIng:
This looks like an elegant solution, many thanks.

Do you have a set of settings or timing to replicate the issue with an empty response? Keeping the request open seems benefitial in any case, but replicating the exact issue with some added debug output might be a good learning, and probably the same affects other part of the code.

Marijn0:
>Do you have a set of settings or timing to replicate the issue with an empty response?

Yes, of course, this can be replicated by:

- Closing all connections to the motionEye server.

- Waiting 10 seconds (or the number of seconds set in motioneye.conf [here](https://github.com/motioneye-project/motioneye/blob/4f1b7b50182a6a23cddf3fa88fb0b40f0e667a20/motioneye/extra/motioneye.conf.sample#L54)) for the mjpg client to close.

- Request a snapshot using: `curl -i http://127.0.0.1:8765/picture/1/current`, and check the `content-length`.
(this should now be 0)

- Try again within 10 seconds and check the `content-length` again. 
(this returns a valid JPEG because the mjpg client was started by the first request).

>Keeping the request open seems benefitial in any case, but replicating the exact issue with some added debug output might be a good learning, and probably the same affects other part of the code.

As far as I can see, this only affects `/picture/<id>/current`, since it relies on retrying while waiting for a frame to become available, and I did not find other handlers using this same retry pattern

Maybe we could add an HTTP error when no picture is available? For example:
```
raise HTTPError(503, f'Failed to get current picture for camera {camera_id}: no frame available',)

MichaIng (approved):
Works perfectly. A `curl` to the API in case also takes a little longer if no image is available yet (the retry loop), so one can "feel" in action what is going on 👍.

PR #3273, opened 2026-02-15
Allow motion's %$ conversion specifier

### LINKED ISSUE #3069: Image filename and movie filename fields don't allow the use of %$
Motion allows the use of %$ conversion specifier as the camera name - including in picture_filename, snapshot_filename and movie_filename. However, Motioneye doesn't allow this - presumably as it thinks it will be passed to the OS as is. Would it be possible to change the code so that Motioneye recognises %$ as a valid character sequence in paths please? I noticed that Motioneye already allows the use of %$ in the following fields: , on_event_start, one_event_end, on_movie_end, on_picture_save

MichaIng:

My only concern was that it re-introduced this RCE vulnerability: https://github.com/motioneye-project/motioneye/security/advisories/GHSA-j945-qm58-4gjx
However, the `%$` seems to be replaced by `motion` before it is piped through the shell, so `%$VARIALE` and `%$(command)` are expanded to `Camera1VARIABLE` and `Camera1(command)` respectively, no variable or command substitution hence happens.


MichaIng:
@copilot 
A translated string is expanded a way here, which does not require human re-translation. Please replace the respective origin string in the backend's `.po` files, and add the `%$` to the translations as well. So `msgmerge` (and Weblate) won't treat this as new string, but keep the old translation just with `%$` added.

MichaIng:
Great #3274 is ready to be merged into whichever PR adds this change. That way we avoid any additional translation needs. A blueprint for whenever we change source strings in a sufficiently trivial way, or in a way which does not render the translations entirely wrong. If some native speaker shall better have a look, we can let Copilot add the `fuzzy` tag, maybe even another custom tag: so it appears as unfinished in Weblate, but without the existing translation lost, and optionally with some custom explanation.

MichaIng:

PR #3282, opened 2026-02-16
chore(frontent): update deprecated jQuery event shorthands

MichaIng:
Basically all event handlers and triggers should be applied via `.on('event', handler)` and `.trigger('event')` respectively. The ones updated here do still work, but are deprecated.


MichaIng:
@copilot 
With these changes, camera frames are not added to the GUI anymore, and `cameraIds.sort()` hence throws an error, since no frames have been found to derive the IDs from. It looks like some migration was wrong, or applied to a non-jQuery object. Please try to find the culprit.

PR #3298, opened 2026-03-01
Make admin username editable in GUI

### LINKED ISSUE #2946: default username admin cannot be changed
### Option to change admin username from admin to custom user defined string.

**Description**:
There is no option to change admin user account, by default it uses admin and is not editable. This is not secure considering we are dealing with cameras as admin is the common way to exploit any system. If we can make it to be renamed to something else, this could be better than having just admin. 

**How to reproduce** : 
We are using docker version of motioneye to setup our instance with ccrisan/motioneye:master-amd64 as image as instruction in documentation page. 

**Expected behaviour**: 
_If there can be a environment variable or config added to change admin, that would be neat to deal with this situation._

Hope this helps. Good Luck!

MichaIng:

### DIFF
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -2298,10 +2298,10 @@ def _dict_to_conf(lines, data, list_names=None):
 def _set_default_motion(data):
     data.setdefault('@enabled', True)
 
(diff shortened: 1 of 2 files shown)


MichaIng:

PR #3300, opened 2026-03-03
Handle .mpg files as movie files

### LINKED ISSUE #2524: need to add .mpg extension to the cleaning task
I believe that here:

https://github.com/motioneye-project/motioneye/blob/b2fbf8aef3f203f7f5c5f8f00d82dfeab14ac863/motioneye/mediafiles.py#L45

Now timelapse video added through the configuration of the camera (`timelapse_codec mpg`) is not removed automatically.

MichaIng:


MichaIng:

PR #3301, opened 2026-03-04
Basic Authentication: Fix base64 decoding method in utils

hajekj:
`base64.decodebytes` expects bytes input, but string is being passed which voids the basic authentication. This PR fixes it and lets Basic authentication to work again.

### DIFF
--- a/motioneye/utils/__init__.py
+++ b/motioneye/utils/__init__.py
@@ -298,7 +298,7 @@ def parse_basic_header(header):
     encoded = parts[1]
 
     try:
-        decoded = base64.decodebytes(encoded)
+        decoded = base64.b64decode(encoded).decode('utf-8')
 
     except:
         return None
--- a/motioneye/utils/__init__.py
+++ b/motioneye/utils/__init__.py
@@ -298,7 +298,7 @@ def parse_basic_header(header):
     encoded = parts[1]
 
     try:
-        decoded = base64.b64decode(encoded).decode('utf-8')
+        decoded = base64.b64decode(encoded).decode()
 
     except:
         return None


motioneye/utils/__init__.py
```diff
@@ -294,7 +294,7 @@ def parse_basic_header(header):
     encoded = parts[1]
 
     try:
-        decoded = base64.decodebytes(encoded)
+        decoded = base64.b64decode(encoded).decode('utf-8')
```
MichaIng: UTF-8 is the default anyway:
```suggestion
        decoded = base64.b64decode(encoded).decode()
```

MichaIng (changes requested):
Thanks, all correct:
- Only used [here](https://github.com/motioneye-project/motioneye/blob/63cbb75/motioneye/handlers/base.py#L123) in the base handler for motionEye user authentication, passing Tornado `HTTPServerRequest.headers['Authorization']` which is a string: https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.request
- And `decodebytes()` not only expects bytes but is also deprecated, superseded by `b64decode` which supports both, strings and byte-like objects: https://docs.python.org/3/library/base64.html

But why did no one (else) face authentication issues so far? Was/is `decodebytes()` maybe just an alias for `b64decode()` for backwards compatibility or so? And why did `decoded.split()` never fail (as `decodebytes()` is documented to also return bytes)?

But let's skip the redundant `'utf-8'` in `decode()`.

Marijn0:
>But why did no one (else) face authentication issues so far?

Someone would need to set `http_basic_auth true` in `motioneye.conf`, right?
I guess not many users have been using that option. 😅

MichaIng:
Ohh, right how I missed that. So this enables an alternative authentication method, not requiring a signature in case (which however still works). What is the use case?
- Reverse proxies which have HTTP authentication set up anyway, so with a matching user, this avoids the need for doubled authentication?
- To make the (motionEye) API easier available, compared to forging the signature query string?

hajekj:
Applied your suggestion.

Honestly, I haven't looked into the signed authentication option. I just enabled basic, so I can ping various API endpoints from uptime-kuma for monitoring. It was easier for me to figure out than the signed authentication. I only found reference implementation here - https://github.com/motioneye-project/motioneye-client/blob/main/motioneye_client/client.py#L102 - and if I understand it correctly, I would have to create a unique signature for every URL (endpoint I am healthchecking) rather than using the same set of credentials (which Basic allows me to do).

Also out of curiosity - what is the idea behind the signed authentication (I understand this is different topic)? Is it because the streams are loaded via GET requests so you need to pass in credentials in some way? I would assume it is similar to signed URLs with various object storages, just with the difference that the signature in MotionEye doesn't have any expiration?

MichaIng (approved):

MichaIng:
Thanks! I did not develop the signature authentication, but that GET requests can be used without passing the plain text password around is of course an argument. And that it is per-URI and per-POST-body hardens it, not so much against attackers (which can read the logic in the frontend and compute the signature themselves ... if they can 😄), but against human or code mistakes: one can be sure that the request with its body goes to the intended URI and not accidentally elsewhere.

So easier API access is your reason. Yeah totally understood, figuring out the method from the code to compute a number of signatures to add to Uptime Kuma for monitoring is a bit annoying. Just keep in mind that this means plain text passwords are sent, if you do not use HTTPS via reverse proxy on the same host machine. Might not be a concern in a trusted and secured home network, of course.

MichaIng:
This really is one of the bugs I wonder how it could have ever happened: I mean even if the code in `get_current_user` was changed, and it did send bytes originally, `header.split()` and `decoded.split()` is used in `parse_basic_header()`, which must have failed before. There were a bunch of other such bugs which look like stuff was implemented but never tested, and slipped through linting as well due to missing type annotations (which we really should start to apply consequently), same with tests: #3303

PR #3254, opened 2026-01-11
Fix Python 3.14 compatibility by moving subprocess function to top level

ddxtanx:
Currently when creating the multiprocessing pool, I'm encountering a pickling error that seems to be because the initialization function is not declared at the top level.
```
Jan 11 00:02:22 ddxarchlinux meyectl[320907]: Traceback (most recent call last):
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:   File "/usr/bin/meyectl", line 7, in <module>
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:     sys.exit(main())
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:              ~~~~^^
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:   File "/usr/lib/python3.14/site-packages/motioneye/meyectl.py", line 337, in main
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:     server.main(arg_parser, sys.argv[2:], command[:-6])
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:     ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:   File "/usr/lib/python3.14/site-packages/motioneye/server.py", line 504, in main
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:     run()
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:     ~~~^^
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:   File "/usr/lib/python3.14/site-packages/motioneye/server.py", line 441, in run
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:     tasks.start()
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:     ~~~~~~~~~~~^^
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:   File "/usr/lib/python3.14/site-packages/motioneye/tasks.py", line 54, in start
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:     _pool = multiprocessing.Pool(_POOL_SIZE, initializer=init_pool_process)
Jan 11 00:02:22 ddxarchlinux meyectl[320907]:   File "/usr/lib/python3.14/multiprocessing/context.py", line 119, in Pool
[...]

PR #3312, opened 2026-03-09
Beta 0.44.0b1

MichaIng:
Whats new:
- S3 uploads do now work with and without path elements in the filename, and the "Include Subdirectories" slider has an actual effect:
  - #2760 @pbouill @MichaIng
- The exception it a remote motionEye instance does not respond to HTTP requests at all is now properly handled:
  - #3079 @Matthew1471
- It is now possible to do S3 uploads with SSE-C server-side encryption enabled, defining the base64-encoded key:
  - #3192 @tuxgasy
- Cameras can now be flagged as admin-only, which hides them from the normal/surveillance user GUI and requires admin credentials to access their API:
  - #3245 @Marijn0
- The `picture/ID/current` API has been fixed for requests which are not done within `mjpg_client_idle_timeout` (default: 10) seconds after another:
  - #3253 @Marijn0
- Python 3.14 compatibility has been fixed:
  - #3254 @ddxtanx
- Accessing camera streaming with digest authentication enabled has been fixed @MichaIng:
  - #3265
  - #3266
- Streaming access is not automatically done via normal/surveillance user credentials anymore, instead one needs to enter username and password for each streamed camera individually. In the same turn, streaming is disabled by default, and needs to be enabled explicitly. On motionEye upgrade, existing camera re-streams are not affected, but we recommend to check whether re-streaming is needed, and if so, to apply individual credentials for each camera:
  - #3268 @Marijn0
- Our CI/CD tests have been hardened and extended, applying certain coding standards and typing across our code more widely:
  - #3271 @MichaIng
- Our frontend went through some major cleanup: removing obsolete scripts, updating all other scripts, updating or jQuery methods, or switching to vanilla JavaScript, and merging some of the frontend input validators to reduce the number of event listeners @MichaIng:
  - #3275
  - #3277
  - #3280
  - #3282
  - #3284
  - #3285
- Passwords and usernames with special characters for RTSP network cameras do not cause authentication issues anymore:
  - #3296 @Marijn0
- API requests via basic authentication with `http_basic_auth true` in `motioneye.conf` do not fail anymore:
  - #3301 @hajekj
- The `motion` configuration files are now created and stored with `0600` UNIX mode, to prevent read access for anyone but the motionEye daemon user and root @MichaIng:
  - #3311
  - #3316
- Any file or directory paths passed via API requests or internally do now cause a 403 response or exception respectively, when containing traversal path elements:
  - #3313 @MichaIng
- Several translation updates, many thanks to all @weblate contributors ❤️

🚨 Since this version contains a number of security-related changes, we highly recommend to update your instances ASAP 🚀.


MichaIng:
Actually why "0.43.2"? This includes a bunch of new features, hence no reason to be shy raising the minor version string => "0.44.0".

Marijn0 (approved):

PR #3313, opened 2026-03-10
enh(handlers): prevent path traversal in API

MichaIng:
It is currently possible to traverse to paths outside of the camera data directory with a number of API requests. This commit prevents path traversal directly at the initial GET/POST API handler methods, as well as complements the restriction in the final media file methods invoked by the handlers, to assure also internal usage of those methods deny path traversal.

Add type definitions to function arguments, and use Pillow default BICUBIC resampling filter instead of BILINEAR for thumbnails: `Image.FILTER` throws some false positive mypy annotation, and indeed `Image.Resampling.FILTER` exists instead, but requires a certain Pillow version, while the other constants is still valid. BICUBIC anyway has the better quality/perfornance ratio, being a reasonable default, and the possibly raised processing time for a preview API request should be negligible.

### DIFF
--- a/motioneye/handlers/movie.py
+++ b/motioneye/handlers/movie.py
@@ -28,6 +28,11 @@ __all__ = ('MovieHandler',)
 
 class MovieHandler(BaseHandler):
     async def get(self, camera_id, op, filename=None):
+        if filename is not None and '..' in filename.split('/'):
+            raise HTTPError(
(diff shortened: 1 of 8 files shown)


MichaIng:
@copilot 
Please create tests to verify that using any path traversal elements as plain or URL-encoded with the affected API endpoints result in a 403 response. Do the same (for plain) traversal elements with the affected mediafiles and remote module methods.

PR #3322, opened 2026-04-05
Use Tornado's parse_cookie to correctly unquote cookie values

### LINKED ISSUE #3082: 0.43.1b2 - Monitor Info is \"\"
On the latest pre-release (0.43.1b2) remote cameras display \\"\\":

image

This is from the set-cookie : `monitor_info_1="\"\"";` and probably the function : https://github.com/motioneye-project/motioneye/blob/72af3711d28175669ca06ab09913b788a5fd8617/motioneye/monitor.py#L33

Note this does not appear on Camera5 as Camera5 is still running MotionEyeOS.

Marijn0:
Replace manual cookie splitting in `parse_cookies()` with Tornado’s [parse_cookie()](https://www.tornadoweb.org/en/stable/_modules/tornado/httputil.html#parse_cookie) so quoted values are unquoted correctly.

This normalizes `monitor_info_1=""` to an empty string and prevents it from being re-escaped to `\"\"` in the frontend.

Fixes #3082

(To test this, you do not necessarily need a separate motionEye instance. You can simply add an existing camera again as a remote motionEye camera.)

I also noticed that the [wiki](https://github.com/motioneye-project/motioneye/wiki/Monitoring-Commands) mentions creating the script in `/data/etc/monitor_1`. Is this correct, or should this be `/etc/motioneye`?

### DIFF
--- a/motioneye/utils/__init__.py
+++ b/motioneye/utils/__init__.py
@@ -32,6 +32,7 @@ from dataclasses import dataclass
 
 from PIL import Image, ImageDraw
 from tornado.concurrent import Future
+from tornado.httputil import parse_cookie
 from tornado.ioloop import IOLoop
 
 from motioneye import settings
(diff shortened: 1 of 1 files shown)


MichaIng (approved):
Great, that makes total sense. Also, we do not need to worry about Tornado versions, since motionEye requires the latest available one, solving some security vulnerabilities.

PR #3338, opened 2026-05-22
enh(config): limit backup/restore to specific files

### LINKED ISSUE #3262: task.pickle in backups
Hi

I noticed that in your motion eye general backups you are including the file "tasks.pickle". I understand that capturig this transitory type of data might be helpful to developers, however I would suggest that from a user perspective that restoring transitory information might be less then ideal.  

Restoring tasks.pickle can resurrect tasks that no longer make sense.
Examples of what could go wrong:
- It tries to upload files to a storage location that no longer exists
- It attempts to process motion events from a camera that isn’t connected
- It resumes a cleanup job on a directory that’s empty
- It retries a failed upload from months ago
- It schedules duplicate tasks because the new system already created fresh ones
None of that is helpful during a clean restore.

Thanks

MichaIng:
Backup and restore only those files which are actually generated by motionEye, and reasonable to restore: `motion.conf`, `camera-*.conf`, and `prefs.json`. `uploadservices.json` contains mostly redundant overridden info (dedicated cleanup needed), and refresh + access tokens for Google and Dropbox OAuth2. Restoring expired or revoked tokens does not make sense. If any is invalid, the OAuth2 authorization loop needs to be done in any case. All other files need to be created on console, manually or via motioneye_init. They do not necessarily need to be writable by the daemon, and should hence be consequently excluded.

In case the config path contains more than 100 files, it was assumed to be "a system-wide config directory", and selective `motion.conf` + `camera-*.conf` were added to the backup already, indicating that these were the major files this feature was intended for. This commit aligns things by doing selective backup in any case, adding the reasonable/harmless `prefs.json` (visual preferences for the GUI), and applying the same file selecting for restoring backups.

For easier OS-agnistic filtering, the native Python `tarfile` module is used instead of the external `tar` command.

### DIFF
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -14,13 +14,16 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import collections
-import datetime
-import glob
 import logging
 import os.path
 import subprocess
+import tarfile
+from collections import OrderedDict
+from datetime import timedelta
 from errno import EEXIST, ENOENT
+from fnmatch import fnmatch
+from glob import glob
+from io import BytesIO
 from os import stat
 from re import match, sub
 from secrets import token_hex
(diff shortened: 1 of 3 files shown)


motioneye/config.py
```diff
@@ -2094,64 +2097,43 @@ def invalidate_monitor_commands():
...
 
     try:
-        p = subprocess.Popen(
-            cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
-        )
-        msg = p.communicate(content)[0]
-        if msg:
-            logging.error(f'failed to restore configuration: {msg}')
-            return False
+        with tarfile.open(fileobj=BytesIO(content)) as tf:
```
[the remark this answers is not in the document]
MichaIng: No, the default `r` mode is an alias for `r:*`, both transparent compression support. `r:` would expect explicitly uncompressed input: https://docs.python.org/3/library/tarfile.html#tarfile.open

And of course I tested it locally, and it works as expected.

motioneye/config.py
```diff
@@ -2094,64 +2097,43 @@ def invalidate_monitor_commands():
...
+            members = [
+                m
+                for m in tf.getmembers()
+                if any(fnmatch(m.name.lstrip('./'), p) for p in patterns)
+            ]
+            tf.extractall(settings.CONF_PATH, members=members)  # nosec B202
```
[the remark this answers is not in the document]
MichaIng: We extract only members which match the defined patterns:
```py
patterns = ['motion.conf', 'camera-*.conf', 'prefs.json']
```
For backwards compatibility, `./motion.conf` etc is accepted as well, but absolute paths, traversal elements, or any such are excluded already. Further checks are redundant, hence the explicit `# nosec B202` to mute bandit.

motioneye/config.py
```diff
@@ -2094,64 +2097,43 @@ def invalidate_monitor_commands():
...
-                glob.glob(os.path.join(settings.CONF_PATH, 'camera-*.conf')),
-            )
-        )
-        try:
-            content = utils.call_subprocess(cmd, cwd=settings.CONF_PATH, encoding=None)
-            logging.debug(f'backup file created ({len(content)} bytes)')
+    files = ['motion.conf']
+    files += glob('camera-*.conf', root_dir=settings.CONF_PATH)
+    # prefs.json does not exist until preferences are changed, hence filter with glob()
+    files += glob('prefs.json', root_dir=settings.CONF_PATH)
```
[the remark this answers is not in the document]
MichaIng: Good point. I'll reverted to using the join + basename loop in a more compact way.

MichaIng:
@copilot+claude-sonnet-4.6
Please add tests for backup and restore, which also verify that only files matching `motion.conf`, `camera-*.conf`, and `prefs.json` are added to and restored from the tarball into the motionEye config directory.

tests/test_config.py
```diff
@@ -0,0 +1,246 @@
...
+            f.write(content)
+
+    def _get_tarball_names(self, data: bytes) -> list:
+        with tarfile.open(fileobj=BytesIO(data)) as tf:
+            return tf.getnames()
+
+    @mock.patch.object(settings, 'CONF_PATH')
+    def test_backup_includes_motion_conf(self, mock_conf_path):
+        mock_conf_path.__str__ = lambda _: self.conf_dir
+        settings.CONF_PATH = self.conf_dir
```
MichaIng: @copilot 
There is quite some extra code to patch `settings.*` values, while they are set the same for every test. Wouldn't it be easier to define them once for all tests? We can patch them once we add new tests which actually require other values.

tests/test_config.py
```diff
@@ -0,0 +1,211 @@
...
+        settings.CONF_PATH = cls._orig_conf_path
+        settings.ENABLE_REBOOT = cls._orig_enable_reboot
+        rmtree(cls.conf_dir)
+
+    def setUp(self):
+        # Clean the config dir before each test
+        for name in os.listdir(self.conf_dir):
+            os.remove(os.path.join(self.conf_dir, name))
+        self._invalidate_patch = mock.patch('motioneye.config.invalidate')
+        self._invalidate_patch.start()
```
MichaIng: @copilot 
What is the purpose of this `_invalidate_patch`?

MichaIng: Ah I see it, our `invalidate()` function to clear the config caches. I misread it as path, with `motioneye.conf` in mind.

MichaIng: Right, doesn't hurt, but not needed in tests either.

Marijn0:
Motion/privacy mask files probably also need to be backed up like before? Maybe monitor scripts and action button scripts too, although I’m less sure about those because restoring executable scripts from a backup may have security implications?

(I’m currently not able to test these changes until tomorrow.) But I think it looks good. 👍

MichaIng:
@copilot+claude-sonnet-4.6
Please fix the failed tests. IMO, `_conf_files()` can be removed, and replaced with direct `os.listdir(self.conf_dir)` calls. There is really not much benefit to wrap this into a function.

MichaIng:
@Marijn0 
Mask files is a good point. Monitor and action button scripts are explicitly excluded, as they cannot be added with the GUI either, and we do not want someone with GUI access to implicitly be able to execute any command by restoring custom scripts.

MichaIng:
@copilot+claude-sonnet-4.6 
As mentioned by Marijn0 above, please add mask files to the patterns of files to backup and restore.  If I see it correctly, it is `mask_*.pgm`: https://github.com/motioneye-project/motioneye/blob/enh/backup/motioneye/utils/__init__.py#L457-L465
Complement tests accordingly.

PR #3342, opened 2026-05-24
Translate working schedule labels for all days

Marijn0:
In the working schedule menu, only "from" and "to" for Monday were translated, while Tuesday through Sunday still used hard-coded English text.

Apply the existing gettext labels to every day so the schedule is translated consistently across the full week.


MichaIng (approved):
A funny oversight, great find!

PR #3347, opened 2026-06-06
Fix saving passwords changed without a keydown event

### LINKED ISSUE #3345: Pi Zero 2W Trixie 64-bit - not remembering entered passwords
As expected, the first time I use the web interface, I'm prompted to enter passwords for admin and user. I do so and hit "Apply". These are somehow not being remembered, because if I get out and close my browser and try to come in again, I get asked to set passwords again. 

This is on an up-to-date system using Trixie (gui version, not lite):
`Linux piz3 6.18.33+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.18.33-1+rpt1 (2026-06-01) aarch64 GNU/Linux`

I see this if I query status from the service:
```
Jun 02 09:09:11 piz3 systemd[1]: Started motioneye.service - motionEye Server.
Jun 02 09:09:18 piz3 libcamerify[66998]: WARNING:root:The admin and/or surveillance user has no password assigned. Please login to >
Jun 02 09:09:18 piz3 libcamerify[66998]: configure_logging cmd motioneye: False
Jun 02 09:09:18 piz3 libcamerify[66998]: configure logging to file: None
Jun 02 09:09:18 piz3 libcamerify[66998]:     INFO: hello! this is motionEye server 0.44.0b2

```

Marijn0:
Fix passwords not being sent to the backend when changed without a `keydown` event.

This fixes an issue where passwords entered through a password manager, browser autofill, or right-click paste could appear to be accepted in the UI but were not actually saved after clicking Apply.

Those input methods can trigger a `change` event without triggering `keydown`. The frontend previously required both events before including password fields in the config payload, so the Apply request could be sent without the updated passwords.

Issue: #3345

### DIFF
--- a/motioneye/static/js/main.js
+++ b/motioneye/static/js/main.js
@@ -3,9 +3,6 @@ var CAMERA_FRAMES_CACHE_LIFETIME = 1000;
 
 var pushConfigs = {};
 var pushConfigReboot = false;
-var adminPasswordChanged = {};
-var normalPasswordChanged = {};
-var streamingPasswordChanged = {};
 var refreshDisabled = {}; /* dictionary indexed by cameraId, tells if refresh is disabled for a given camera */
 var singleViewCameraId = null;
 var fullScreenMode = false;
@@ -579,26 +576,6 @@ function initUI() {
             $minimizeSpan.removeClass('open');
         }
     }
-
-    /* update password changed flags */
-    $('#adminPasswordEntry').on('keydown', function () {
-        adminPasswordChanged.keydown = true;
-    });
-    $('#adminPasswordEntry').on('change', function () {
-        adminPasswordChanged.change = true;
-    });
-    $('#normalPasswordEntry').on('keydown', function () {
-        normalPasswordChanged.keydown = true;
-    });
-    $('#normalPasswordEntry').on('change', function () {
-        normalPasswordChanged.change = true;
-    });
-    $('#streamingPasswordEntry').on('keydown', function () {
-        streamingPasswordChanged.keydown = true;
-    });
-    $('#streamingPasswordEntry').on('change', function () {
-        streamingPasswordChanged.change = true;
-    });
 
     /* ui elements that enable/disable other ui elements */
     $('#storageDeviceSelect').on('change', updateConfigUI);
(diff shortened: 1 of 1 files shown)


MichaIng:
I do not understand either why the keydown event was an additional condition.

Actually, if we pre-fill the fields with `*****`, we do not even need to track the "change" event, but take the value as only condition. The latter implies the first, hence the first is redundant, if I am not missing something.

Marijn0:
>Actually, if we pre-fill the fields with `*****`, we do not even need to track the "change" event, but take the value as only condition. The latter implies the first, hence the first is redundant, if I am not missing something.

Good point, so we can just use something like:

```js
if ($('#adminPasswordEntry').val() !== '*****') {
```

which is much cleaner.

I am not sure why all those conditions were needed in the past. Maybe they were there to avoid submitting empty password fields. But now that the admin and surveillance user passwords are required, those extra conditions can be removed AFAICS.

MichaIng:
Even a change + keydown event does not assure the field is not empty. I mean making it empty requires a change and a keydown (assuming no password manager can enter an empty password).

Maybe the conditions were incrementally appended ... yeah, see here: https://github.com/motioneye-project/motioneye/commit/818cd28

So the keydown condition was explicitly added to prevent saving autofilled password, not considering password managers, but browser autocompletion features which sometimes touch all input fields without actually changing the content.

The failsafe "addition" of checking `*****` however made the other unreliable conditions obsolete.

This also explains why there was/is another `*****` check in the backend. I guess this was added when it was not present in the frontend yet. Now that it is, the backend does not need to filter this anymore.

Marijn0:
Good find!

> This also explains why there was/is another `*****` check in the backend. I guess this was added when it was not present in the frontend yet. Now that it is, the backend does not need to filter this anymore.

If there are still any such checks in the backend, could you please point me to them?

MichaIng (approved):
Great when a bug can be fixed by removing and simplifying code 👍🙂

MichaIng:
> If there are still any such checks in the backend, could you please point me to them?

It is not there anymore. I just remember this popped up in a PR, maybe got removed, or maybe someone wanted to add it in the backend. I just remember that I verified the frontend does the `*****` filtering already, and that hence it wouldn't be needed in the backend.

PR #3350, opened 2026-06-08
Fix no-camera placeholder image aspect ratio

Marijn0:
The no-camera SVG placeholder was being squashed because the inline `height="16"` attribute kept the height fixed while the CSS rule `width: 30%` scaled the width freely.
Removed the inline `width="16"` and `height="16"` attributes. By only declaring `width: 30%` in CSS, the aspect ratio is preserved automatically.
https://github.com/motioneye-project/motioneye/blob/2b7327a1334e185b3a837ec0949f367a2f17d7be/motioneye/static/css/main.css#L1274-L1277

<table>
<tr>
<td align="center">Before</td>
<td align="center">After</td>
</tr>
<tr>
<td>
before
</td>
<td>
after
</td>
</tr>
</table>

### DIFF
--- a/motioneye/static/js/main.js
+++ b/motioneye/static/js/main.js
@@ -4634,7 +4634,7 @@ function addCameraFrameUi(cameraConfig) {
     var cameraFrameDiv = $(
             '<div class="camera-frame">' +
                 '<div class="camera-container">' +
-                    '<div class="camera-placeholder"><img class="no-camera" src="' + staticPath + 'img/no-camera.svg" width=16 height=16></div>' +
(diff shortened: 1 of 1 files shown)


MichaIng (approved):
Many thanks for picking this up ❤️!

The icon is a direct child of the camera placeholder, so the 30% refers to that, which seems fine in all real cases. We could turn it into a `max-width: 30%`, so it does not get squashed for extreme wide aspect ratios. Similarly, the fixed 20% top-margin could be replaced with a `vertical-align: middle`, I guess (probably needed on the `camera-placeholder` rather than the img itself).

But this all will hardly ever play a role in practice, great to have this icon finally in correct aspect ratio.

PR #3304, opened 2026-03-05
Fix movie preview generation for movies shorter than 4 seconds

MichaIng:
This commit was done with an anonymous account in context of a study. The actual author might want to open a PR with its true account for proper credits, after the study ended. However, I am opening this here to show that this issue has been addressed, so no one starts doubled work 😉.

PR #3355, opened 2026-06-17
Beta 0.44.0b3

MichaIng:
## What's new
- A new button and dialog was added to change credentials for network cameras, including simple MJPEG cameras, and remote motionEye cameras. It is hence not needed anymore to remove and re-add a camera for this. If the connection to a remote motionEye camera fails, the dialog pops up automatically, which simplifies applying the new remote/client secret, required since motionEye 0.44.0b2: #3353 @Marijn0
- An issue has been solved where passwords changed in the settings panel without manually typing into the input box, e.g. by letting a password manager auto-fill it, were not applied: #3347 @Marijn0
- The aspect ratio of the camera placeholder icon has been fixed: #3350 @Marijn0
- The name of the admin user can now be changed in the settings panel: #3298 @strawberry-poodle (anonymous study account)
- The `motion` `%$` [conversion specifier](https://motion-project.github.io/motion_config.html#conversion_specifiers) to reference the camera name can now be used in media filenames: #3273 @strawberry-poodle (anonymous study account)
- When restoring backups, the file type and paths are now checked more strictly, to assure that only real files are extracted to the root of the config dir, and nowhere else: #3341 @MichaIng
- `.mpg` files inside the camera media directories are not treated as movie files regarding preview generation, cleanup, and similar tasks. Available motionEye options do not currently generate those files, but with custom settings, manual conversions, media files copied from elsewhere, the situation is generally possible: #3300 @strawberry-poodle (anonymous study account)
- Preview images for movie files with less than 4 seconds duration are now generated: #3304 @strawberry-poodle (anonymous study account)
- As always, several translations have been updated. Many thanks to all @weblate contributors! ❤️


MichaIng:
I do not aim to wait long with the stable 0.44.0 release. The intermediate beta is mainly to test the automatic PyPI releases, which should be triggered when we push the 0.44.0b3 pre-release tag.

In the meantime we have CVEs for all fixed security vulnerabilities, hence can publish them as well along with the stable release.

You may skip the language files for reviews. The diffs look larger than they are, due to sorted frontend POT file entries, and Weblate forced that sorting onto the PO files, and our workflow again onto the JSON 😄.
Could make actually sense to switch to ttag or so: https://ttag.js.org/
It seems to be able to translate based on the PO files directly, hence would not require the JSONs at all. One less step in our workflows, and 34 fewer file changes whenever we add/change translated strings in the JavaScript files.

Marijn0:
image

The button doesn’t seem to scale well with text of varying lengths across different languages. It’s not a deal-breaker, but it could be improved. 

FYI: Something I noticed while going through the code recently: `PrefsHandler` allows unauthenticated users to read and write preferences, limited to preferences stored under anonymous. 
https://github.com/motioneye-project/motioneye/blob/358fb6df047a097e8670166635efa32ef6ad33de/motioneye/handlers/base.py#L173-L177 AFAICS, there is no reason to allow this. It’s not a major issue, but it could be improved before the 0.44.0 release. (Perhaps something Copilot can handle 😅).

Going through the diff, it looks good AFAICS.

Marijn0 (approved):

MichaIng:
> The button doesn’t seem to scale well

They have a fixed width: https://github.com/motioneye-project/motioneye/blob/dev/motioneye/static/css/main.css#L381
Could be turned into a `min-width`, so those with longer text adjust. But there is no padding either:

{D3B037E9-9441-4B82-85BA-A7DA0FB7CC85}

Additionally, or alternatively, we could just add proper padding and remove the `width`:

{E85AA65A-298D-43C5-8089-0120240084E4}

Hmm, I think both is best:

{EAD2DBC0-03DF-496C-9BD9-0B9BA5CD0EB4}

{6AE27A6E-9078-4163-9BC8-96C6EB33C14F}

_7em = 8em - 1em passing_

_________

> `PrefsHandler` allows unauthenticated users to read and save preferences

Hmm, I guess the reason was that previously, without a normal user password, embedded camera frames were supposed to show up without authentication? But they do not make use of any of the preferences ... although, the framerate dimmer/factor? ... nope, prefs are not loaded at all for the frame: https://github.com/motioneye-project/motioneye/blob/dev/motioneye/static/js/main.js#L5529-L5531
Also verified via browser dev tools. Let's remove that anonymous prefs load/safe.

MichaIng:
I accidentally raised the minimum Python version to 3.10 using `type1 | type2 | ...` syntax for typing, during 0.44.0 dev cycle. I thought some pre-commit test would prevent this, but it only has Python 3.7 as lower limit for syntax upgrades, no upper limit. Will switch to `Union[type1, type2, ...]` syntax tomorrow, and add a Python 3.7 job to our test matrix. Those are 3.10+ as well after the Ubuntu 20.04 runner was removed by GitHub.

We can raise the requirements to Python 3.8 or 3.9 IMO, with next release. But for 3.10 it is probably a bit early.

PR #3357, opened 2026-06-18
handlers: require authentication for prefs API

MichaIng:
This API is harmless, affects visual preferences only: camera frame alignment and sizes, as well as lowering the displayed camera resolution and framerate.

However, this matters only for the GUI after authentication, hence there is no point to allow using this API without authentication, messing with users' preferences.

For clarity, the functions for getting/setting prefs have been moved into the prefs handler script, and inlined, since they are used only ín this single place.

Since I am lazy to create two PRs, this commit additionally allows the blue settings panel buttons to grow with long text. With some translations, they can be too small. The fixed width is turned into a minimal width with a padding. Buttons with short texts will have the same size, but buttons with longer text will grow with the text length.

{8CF2F25C-9413-4889-982D-ACF2BA2288AE}

### DIFF
--- a/motioneye/handlers/base.py
+++ b/motioneye/handlers/base.py
@@ -22,7 +22,7 @@ from time import time
 
 from tornado.web import HTTPError, RequestHandler
 
-from motioneye import config, prefs, settings, template, utils
+from motioneye import config, settings, template, utils
 from motioneye.utils.authstate import verify_hmac_signature
 
 __all__ = ('BaseHandler', 'NotFoundHandler', 'ManifestHandler')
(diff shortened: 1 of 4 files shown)


Marijn0 (approved):
Looks good, thanks!

Marijn0:
> messing with users' preferences.

For clarity, I believe this was not actually possible:

```python
return prefs.set(self.current_user or 'anonymous', key, value)
```

`self.current_user` was only set after successful authentication. Without authentication, `self.current_user` was `None`, so it fell back to `anonymous`, meaning an unauthenticated request could not modify an authenticated user's preferences.

Since anonymous preferences were not actually used, allowing unauthenticated access served no purpose and could safely be removed.

PR #3360, opened 2026-06-19
fix: restore Python 3.7 compatibility

MichaIng:
by using Python 3.7 compatible typing. Also use f-strings where possible in touched files.

Add Python 3.7 to our test matrix, as pre-commit default version, and add ruff linter and formatter with Python 3.7 as target version.

Also add Ubuntu 26.04 Resolute runners to test matrix for Python 3.14 tests, which are in preview state.

### DIFF
--- a/.github/workflows/test_python.yml
+++ b/.github/workflows/test_python.yml
@@ -15,11 +15,17 @@ jobs:
     strategy:
       matrix:
         include:
+          # Tests with oldest Python version able to build the project on oldest available runner images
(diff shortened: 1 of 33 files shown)


MichaIng:
Finally addressed a bunch of unrelated linter annotations 😅. Anyway, nice cleanup of unused imports, variables, and all such. I guess we can include most flake8 checks now, which overlap with ruff. But something for next release.

PR #3362, opened 2026-06-19
Release 0.44.0

MichaIng:
### Breaking changes: #3332 @motioneye-project
- The signature authentication method has been replaced with session cookies. For using the API, a session cookie needs to be obtained by doing a POST request to the `/login` API, with `username` and `password` arguments as POST data/parameters. If credentials were correct, a session cookie is returned, which needs to be sent with the actual API requests. The cookie has a 24h lifetime, hence scripts/programs should automatically redo the login loop when receiving a 403.
- For remote motionEye cameras exclusively, a HMAC authentication method has been added, generating a signature from a dedicated secret, request URL, nonce, and tight timestamp. This secret is automatically generated for every motionEye instance, and needs to be entered when adding a remote motionEye camera. Implicitly, motionEye v0.44.0 instances cannot work with older versions vice versa. On first admin login after the update, respectively when selecting the remote motionEye camera from the settings panel, a prompt will allow you to enter the new secret. A new button in the settings panel allows to change it any time later, in case it changes at the remote motionEye instance.
- The HTTP basic authentication option has been removed.
- Passwords are now enforced for both users: admin and surveillance/normal, and stored as Argon2 hash on the server. Existing SHA1 hashes and plain text passwords are still functional, and migrated automatically on login. Users are asked to apply passwords on login, if none applied yet.
- The `/action` API does now require authentication.

### Other changes
- S3 uploads do now work with and without path elements in the filename, and the "Include Subdirectories" slider has an actual effect:
  - #2760 @pbouill @MichaIng
- The exception it a remote motionEye instance does not respond to HTTP requests at all is now properly handled:
  - #3079 @Matthew1471
- It is now possible to do S3 uploads with SSE-C server-side encryption enabled, defining the base64-encoded key:
  - #3192 @tuxgasy
- Cameras can now be flagged as admin-only, which hides them from the normal/surveillance user GUI and requires admin credentials to access their API:
  - #3245 @Marijn0
- The `picture/ID/current` API has been fixed for requests which are not done within `mjpg_client_idle_timeout` (default: 10) seconds after another:
  - #3253 @Marijn0
- Python 3.14 compatibility has been fixed:
  - #3254 @ddxtanx
- Accessing camera streaming with digest authentication enabled has been fixed @MichaIng:
  - #3265
  - #3266
- Streaming access is not automatically done via normal/surveillance user credentials anymore, instead one needs to enter username and password for each streamed camera individually. In the same turn, streaming is disabled by default, and needs to be enabled explicitly. On motionEye upgrade, existing camera re-streams are not affected, but we recommend to check whether re-streaming is needed, and if so, to apply individual credentials for each camera:
  - #3268 @Marijn0
- Our CI/CD tests have been hardened and extended, applying certain coding standards and typing across our code more widely:
  - #3271 @MichaIng
- Our frontend went through some major cleanup: removing obsolete scripts, updating all other scripts, updating or jQuery methods, or switching to vanilla JavaScript, and merging some of the frontend input validators to reduce the number of event listeners @MichaIng:
  - #3275
  - #3277
  - #3280
  - #3282
  - #3284
  - #3285
- Passwords and usernames with special characters for RTSP network cameras do not cause authentication issues anymore:
  - #3296 @Marijn0
- The `motion` configuration files are now created and stored with `0600` UNIX mode, to prevent read access for anyone but the motionEye daemon user and root @MichaIng:
  - #3311
  - #3316
- Any file or directory paths passed via API requests or internally do now cause a 403 response or exception respectively, when containing traversal path elements:
  - #3313 @MichaIng
- Fixed monitor info text on camera frames being wrapped into double quotes with extra backslash character `\"\"`:
  - #3322 @Marijn0 
- The input path validation at the backend has been extended, to prevent any kind of escaping the camera data dir:
  - #3324 @MichaIng
- The backup/restore feature is now limited to those config files which are actually generated by the web interface: `motion.conf`, `camera-*.conf`, `prefs.json`, and `mask-*.pgm`. Files that need to be created from console, like `motioneye.conf` and camera action scripts, are not included, and won't be restored anymore, if present in old backups:
  - #3338 @MichaIng
- A new button and dialog was added to change credentials for network cameras, including simple MJPEG cameras, and remote motionEye cameras. It is hence not needed anymore to remove and re-add a camera for this. If the connection to a remote motionEye camera fails, the dialog pops up automatically, which simplifies applying the new remote/client secret, required since motionEye 0.44.0b2:
  - #3353 @Marijn0
- An issue has been solved where passwords changed in the settings panel without manually typing into the input box, e.g. by letting a password manager auto-fill it, were not applied:
  - #3347 @Marijn0
- The aspect ratio of the camera placeholder icon has been fixed:
  - #3350 @Marijn0
- The name of the admin user can now be changed in the settings panel:
  - #3298 @strawberry-poodle (anonymous study account)
- The `motion` `%$` [conversion specifier](https://motion-project.github.io/motion_config.html#conversion_specifiers) to reference the camera name can now be used in media filenames:
  - #3273 @strawberry-poodle (anonymous study account)
- When restoring backups, the file type and paths are now checked more strictly, to assure that only real files are extracted to the root of the config dir, and nowhere else:
  - #3341 @MichaIng
- `.mpg` files inside the camera media directories are not treated as movie files regarding preview generation, cleanup, and similar tasks. Available motionEye options do not currently generate those files, but with custom settings, manual conversions, media files copied from elsewhere, the situation is generally possible:
  - #3300 @strawberry-poodle (anonymous study account)
- Preview images for movie files with less than 4 seconds duration are now generated:
  - #3304 @strawberry-poodle (anonymous study account)
- Several translations have been updated, and an initial addition of the Bulgarian language was done. Many thanks to all @weblate contributors! ❤️

🚨 Since this version contains a number of security-related changes, we highly recommend to update your instances ASAP 🚀.


motioneye/handlers/login.py
```diff
@@ -15,17 +15,143 @@
...
+                main_config.get('@admin_username')
+                if user == 'admin'
+                else main_config.get('@normal_username')
+            )
+            self.finish_json({'user': user, 'username': username or user})
+        else:
+            self.set_status(401)
+            self.finish_json({'error': 'not authenticated'})
+
+    def post(self) -> None:
```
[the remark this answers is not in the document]
MichaIng: Right, that would be cleaner, and I applied this in at least one case already. However, `finish_json()` -> tornado's `finish()` returns always None, and we have mypy in CI which checks this. Hence no action needed as long as mypy is happy.

Marijn0:
In the release notes it states:

>- API requests via basic authentication with `http_basic_auth true` in `motioneye.conf` do not fail anymore:
>  - #3301

Since the HTTP basic authentication option was later removed during the 0.44.0 beta cycle, this is no longer relevant for the final release IMO.

MichaIng:
Good find, that might have been confusing 😄. Btw: I'll add the CVE/advisory references later on in https://github.com/motioneye-project/motioneye/releases, once PyPI and Docker Hub builds were published.

Marijn0:
And the version string needs to be updated as well. 🙂

MichaIng:
Yes, my last unreviewed file, to change right before merge 😄.

EDIT: Though, not much reason to wait with that. Just done.

motioneye/static/js/main.js
```diff
@@ -3580,70 +3487,108 @@ function runConfirmDialog(message, onYes, options) {
...
+                                runAlertDialog(
+                                    i18n.gettext('Please set a password for your account.'),
+                                    function () {
+                                        openSettings();
+                                        setTimeout(function () {
+                                            $('#adminPasswordEntry').focus();
+                                        }, 500);
+                                    }
+                                );
+                                return;
```
Marijn0, line 3570: Some users may have only set a password for the admin account and never for
the normal user. When an admin logs in, we only show a message if the admin
itself has no password set. In hindsight, I would have preferred to also
show a warning when the normal user has no password, but this would require
backend changes as well: login.py (~L152) would need to include a
'normal_password_missing' flag in the admin login response, and
forcePasswordChange would need to be extended to cover this case.

MichaIng: At least if the normal user logs in, it gets the prompt. And if the admin saves settings for any reason, it will be forced to enter a normal user password as well. So this additional flag is relevant for updated instances only, if the admin has a password already but the normal user not, and only until settings are saved for any reason, or the normal user logs in. I think it is okay.

Marijn0 (approved):

zagrim (approved):
👍

PR #3372, opened 2026-06-27
enh(config): use "pause" setting from motion 4.4 on

### LINKED ISSUE #3371: Switching "Still images" to "off" and "on" again changes "capture mode" to "manual"
Motion: 4.7.1
MotionEye: 0.44

Switching "Still images" to "off" and "on" again changes "capture mode" to "manual".
Having set "capture mode" to "motion triggered" press "apply".

Image

Then set "still images" to "off". press "apply"
Next set "still images" to "on" and observe "capture mode" changed to "manual"

Image

MichaIng:
This allows starting motion with motion detection disabled, rendering the API request at startup obsolete. That request could have been done too early if motion took more than 2 seconds to start.

### DIFF
--- a/motioneye/motionctl.py
+++ b/motioneye/motionctl.py
@@ -134,7 +134,9 @@ def start(deferred=False):
     with open(motion_pid_path, 'w') as f:
         f.write(str(pid) + '\n')
 
-    IOLoop.current().spawn_callback(_disable_initial_motion_detection)
+    # disable motion detection if version < 4.4, otherwise the new "pause" setting does
+    if not is_motion_post43():
+        IOLoop.current().spawn_callback(_disable_initial_motion_detection)
 
     # if mjpg client idle timeout is disabled, create mjpg clients for all cameras by default
     if not settings.MJPG_CLIENT_IDLE_TIMEOUT:
@@ -372,8 +374,6 @@ def has_h264_omx_support():
     if not binary:
         return False
 
-    # TODO also check for motion codec parameter support
-
     return 'h264_omx' in codecs.get('h264', {}).get('encoders', set())
 
 
@@ -381,8 +381,6 @@ def has_h264_v4l2m2m_support():
     binary, version, codecs = mediafiles.find_ffmpeg()
     if not binary:
         return False
-
-    # TODO also check for motion codec parameter support
 
     return 'h264_v4l2m2m' in codecs.get('h264', {}).get('encoders', set())
 
@@ -436,7 +434,7 @@ def has_hevc_qsv_support():
 
 
 def resolution_is_valid(width, height):
-    # width & height must be be modulo 8
+    # width & height must be modulo 8
 
     if width % 8:
         return False
--- a/motioneye/motionctl.py
+++ b/motioneye/motionctl.py
@@ -14,14 +14,15 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import errno
 import logging
-import os.path
 import re
-import signal
-import subprocess
-import time
+from errno import ECHILD, ESRCH
+from os import WNOHANG, kill, waitpid
+from os.path import exists, join
 from shlex import quote
+from signal import SIGKILL, SIGTERM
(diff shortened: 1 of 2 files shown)


motioneye/motionctl.py
```diff
@@ -134,7 +134,9 @@ def start(deferred=False):
     with open(motion_pid_path, 'w') as f:
         f.write(str(pid) + '\n')
 
-    IOLoop.current().spawn_callback(_disable_initial_motion_detection)
+    # disable motion detection if version < 4.4, otherwise the new "pause" setting does
+    if not is_motion_post43():
+        IOLoop.current().spawn_callback(_disable_initial_motion_detection)
```
Marijn0, line 140: The new `pause` config option is only written to the camera config when settings are applied. So after an upgrade, an existing camera with `@motion_detection = False` may still start with motion detection enabled until the camera settings are saved again.

MichaIng: Yes, first of all I want to see whether this works and solves the issue. Works fine here, and one can see how motion detection is disabled immediately for those cameras which have the setting, and 2 seconds later for the others (if disabled in GUI).

Not yet sure how to do the migration. Loading every camera config to check whether `pause` has been applied already is quite a big overhead. A config version string in the main config would be handy, as this is loaded in `meyectl startserver` anyway, and could then apply specific migrations.
EDIT: Actually we could implement it now, one for the motionEye version, and one for the motion version at which configs were last saved. The latter could be also used to automate what is currently in this stray script https://github.com/motioneye-project/motioneye/blob/dev/motioneye/scripts/migrateconf.sh, surely handy when motion v5 is released and once motionEye supports it.

Maybe it is fine to document that the config needs to be saved once after upgrades, for cameras with disabled motion detection. If it matters, people will recognize, and usually attempt to solve it by toggling setting anyway?

Marijn0: Yes, I think that is acceptable IMO.

We should probably add the new `pause` config option to `_set_default_motion_camera()` and `_USED_MOTION_OPTIONS`, right?

Otherwise, if I remember correctly, the option can end up at the bottom of the config because it is not treated as a known Motion option.

MichaIng: Right, it ends up at the bottom right now. I'll add it to `_USED_MOTION_OPTIONS`, but adding it to `_set_default_motion_camera` means that it is applied for motion < 4.4 as well. We'd need to actively unset it again, else old motion throws an error on startup.

MichaIng: Hmm, even when in `_USED_MOTION_OPTIONS` and a default applied in `_set_default_motion_camera()` (which I am testing locally), it ends up at the bottom. Only for newly added cameras it is sorted based on when it was declared, i.e. where it is located in `_set_default_motion_camera()`.

In `_USED_MOTION_OPTIONS` it does not currently have any effect, as this is only relevant when moving values from UI object to internal dictionary, vice versa. But `pause` is not toggled in the UI directly, but added as inversed value of `@motion_detection` just before the config file is written.

Actually, since it currently is such a special config applied only on certain motion versions and not handled/transferred internally anywhere, I think it is fine to not apply a default value and not have it in `_USED_MOTION_OPTIONS` for now. We can implement it properly as replacement for `@motion_detection` when dropping support for motion < 4.4.

Marijn0: I remember messing with this in #3268. Looks like I forgot to add `stream_authentication` to `_set_default_motion_camera` too. 😬😅

Marijn0 (approved):

MichaIng:
Okay, let's go with this. We can add a config file versioning later, along with conditional config file update/rewrite if motionEye and/or motion versions changed a certain way. It is a convenience feature mostly.

Marijn0:
FYI: the new `pause` option appears in the `Extra Options` field in the GUI. 
It's something we should look into when time allows. 😃

MichaIng:
Hmm, so `_USED_MOTION_OPTIONS` does have an effect after all. Makes sense. I saw that it is only for translating back and forth UI to dict values, but did not think about the fact that unknown settings are still passed and end in Extra Options.

PR #3374, opened 2026-06-28
feat(upload): add "remove files after upload" option

### LINKED ISSUE #3089: Clearing up moves sooner than 1 day
I've just installed Motioneye via Docker on Raspberry Pi 5 via the following docker compose file. My SD card is only 32GB and I only want to upload my recordings to my Synology NAS. I don't need them on the Pi. Since there's no option to set Network Share because it relies on additional packages and root user, I've turned on the Upload Service which is working, however, there's no point keeping media files on the pi beyond say 15 mins. Can i set Preserve Movies to custom and set a float for movies lifetime, for instance, 0.005 or something? What is this value?

```yml
services:
  motioneye:
    image: ghcr.io/motioneye-project/motioneye:edge
    container_name: motioneye
    ports:
      - "8765:8765"
      - "9081:9081"
    volumes:
      - ./motioneye:/etc/motioneye
      - /etc/localtime:/etc/localtime:ro
      - /data/motioneye:/var/lib/motioneye
    devices:
      - /dev/video0:/dev/video0
    restart: unless-stopped
```

JamBalaya56562:
## What

Adds a per-camera **"Remove Files After Upload"** option (`@clean_uploaded`) under **Storage**. When enabled, a movie/picture is deleted from local storage immediately after the upload service has **successfully** uploaded it, freeing space on small media such as SD cards.

Implements the feature @MichaIng proposed in #3089.

## Why

#3089: the only local-retention control today is "Preserve Movies/Pictures", which is whole-days only (sub-day values round to 0 = "keep forever"), so it can't promptly free a small SD card after media is uploaded to a NAS/cloud. This option does exactly that.

## How

- **`config.py`** — new `@clean_uploaded` boolean, mirroring `@clean_cloud_enabled` (ui↔dict mapping + default `False`).
- **`handlers/relay_event.py`** — thread `media_type` (`'movie'`/`'picture'`) into the upload task.
- **`uploadservices.upload_media_file`** — on a successful `upload_file` (no exception), when `@clean_uploaded` is set, remove the local file via `mediafiles.del_media_content` (file + `.thumb` + now-empty date dirs). A **failed upload keeps the file**. A function-local import avoids the `mediafiles`/`config` ↔ `uploadservices` circular import.
- **`mediafiles.cleanup_media`** — skip cloud cleanup when `@clean_uploaded` is set: the two options are mutually exclusive, because once local files are removed after upload, `clean_cloud` would treat the just-uploaded cloud copies as orphaned and delete them.
- **`templates/main.html` / `static/js/main.js`** — new Storage checkbox (shown for all upload services), made mutually exclusive in the UI with the cloud-cleanup switch.

## Notes / scope

- Per-type behaviour already comes from `@upload_movie` / `@upload_picture`, which decide what gets uploaded; a single boolean is sufficient since a deleted file is always one that was uploaded.
- On `movie_end`, the thumbnail (`make_movie_preview`) and the upload are both scheduled at +5s in the single-worker task pool, with the preview enqueued first; preview failures are already non-fatal. A stronger ordering guarantee is out of scope.
- Single-file **cloud** deletion sync is unchanged (the existing `clean_cloud` stays directory-oriented).
- New UI strings follow the existing `{{ _("...") }}` Esperanto source-string convention; wording can be adjusted as preferred.

## Testing

- Adds `tests/test_uploadservices.py` covering the new cleanup behaviour: deletes on a successful upload when enabled, keeps the file when disabled or when the upload fails, and never lets a cleanup error propagate.
- Full `pytest` suite runs green (existing + new) in a Linux container: **121 passed, 115 subtests passed**.
- `ruff check` + `ruff format --check` pass on all changed `.py` files; `node --check` passes for `main.js`.

Closes #3089

### DIFF
--- a/motioneye/static/js/main.js
+++ b/motioneye/static/js/main.js
@@ -695,8 +695,18 @@ function initUI() {
         var folder = $('#uploadLocationEntry').val();
         console.log('cleanCloudEnabled', enabled, folder);
         if (enabled) {
+            /* mutually exclusive with "remove files after upload" */
+            $('#cleanUploadedSwitch')[0].checked = false;
             runAlertDialog(( i18n.gettext('Ĉi rekursie forigos ĉiujn dosierojn ĉeestantajn en la nuba dosierujo "') + folder +
                     i18n.gettext('", ne nur tiuj alŝutitaj de motionEye!')));
+        }
+    });
+    $('#cleanUploadedSwitch').on('change', function () {
+        if (this.checked && $('#cleanCloudEnabledSwitch')[0].checked) {
+            /* mutually exclusive with cloud cleanup, which would delete the
+             * just-uploaded copies once the local files are gone */
+            $('#cleanCloudEnabledSwitch')[0].checked = false;
+            runAlertDialog(i18n.gettext('"Purigi la nubon" estis malŝaltita ĉar ĝi konfliktas kun "Forigi Dosierojn Post Alŝuto".'));
         }
     });
 
@@ -1869,6 +1879,7 @@ function cameraUi2Dict() {
         'upload_bucket': $('#uploadBucketEntry').val(),
         'upload_sse_c_key': $('#uploadSseCKeyEntry').val(),
         'clean_cloud_enabled': $('#cleanCloudEnabledSwitch')[0].checked,
+        'clean_uploaded': $('#cleanUploadedSwitch')[0].checked,
         'web_hook_storage_enabled': $('#webHookStorageEnabledSwitch')[0].checked,
         'web_hook_storage_url': $('#webHookStorageUrlEntry').val(),
         'web_hook_storage_http_method': $('#webHookStorageHttpMethodSelect').val(),
@@ -2213,6 +2224,7 @@ function dict2CameraUi(dict) {
     $('#uploadBucketEntry').val(dict['upload_bucket']); markHideIfNull('upload_bucket', 'uploadBucketEntry');
     $('#uploadSseCKeyEntry').val(dict['upload_sse_c_key']); markHideIfNull('upload_sse_c_key', 'uploadSseCKeyEntry');
     $('#cleanCloudEnabledSwitch')[0].checked = dict['clean_cloud_enabled']; markHideIfNull('clean_cloud_enabled', 'cleanCloudEnabledSwitch');
+    $('#cleanUploadedSwitch')[0].checked = dict['clean_uploaded']; markHideIfNull('clean_uploaded', 'cleanUploadedSwitch');
 
     $('#webHookStorageEnabledSwitch')[0].checked = dict['web_hook_storage_enabled']; markHideIfNull('web_hook_storage_enabled', 'webHookStorageEnabledSwitch');
     $('#webHookStorageUrlEntry').val(dict['web_hook_storage_url']);
--- a/motioneye/static/js/main.js
+++ b/motioneye/static/js/main.js
@@ -695,8 +695,9 @@ function initUI() {
         var folder = $('#uploadLocationEntry').val();
         console.log('cleanCloudEnabled', enabled, folder);
         if (enabled) {
-            /* mutually exclusive with "remove files after upload" */
-            $('#cleanUploadedSwitch')[0].checked = false;
+            /* mutually exclusive with "remove files after upload"; trigger
+             * change so the styled switch updates visually, not just .checked */
+            $('#cleanUploadedSwitch').prop('checked', false).trigger('change');
             runAlertDialog(( i18n.gettext('Ĉi rekursie forigos ĉiujn dosierojn ĉeestantajn en la nuba dosierujo "') + folder +
                     i18n.gettext('", ne nur tiuj alŝutitaj de motionEye!')));
         }
@@ -704,8 +705,9 @@ function initUI() {
     $('#cleanUploadedSwitch').on('change', function () {
         if (this.checked && $('#cleanCloudEnabledSwitch')[0].checked) {
             /* mutually exclusive with cloud cleanup, which would delete the
-             * just-uploaded copies once the local files are gone */
-            $('#cleanCloudEnabledSwitch')[0].checked = false;
+             * just-uploaded copies once the local files are gone; trigger
+             * change so the styled switch updates visually, not just .checked */
+            $('#cleanCloudEnabledSwitch').prop('checked', false).trigger('change');
             runAlertDialog(i18n.gettext('"Purigi la nubon" estis malŝaltita ĉar ĝi konfliktas kun "Forigi Dosierojn Post Alŝuto".'));
         }
     });
--- a/motioneye/uploadservices.py
+++ b/motioneye/uploadservices.py
@@ -1319,7 +1319,15 @@ def update(camera_id, service_name, settings):
     service.save()
 
 
-def upload_media_file(camera_id, camera_name, target_dir, service_name, filename):
+def upload_media_file(
+    camera_id,
+    camera_name,
+    target_dir,
+    service_name,
+    filename,
+    media_type=None,
+    clean_uploaded=False,
+):
     service = get(camera_id, service_name)
     if not service:
         return logging.error(
@@ -1332,6 +1340,30 @@ def upload_media_file(camera_id, camera_name, target_dir, service_name, filename
     except Exception as e:
         logging.error(
             f'failed to upload file "{filename}" with service {service}: {e}',
+            exc_info=True,
+        )
+        return  # upload failed: keep the local file
+
+    # the caller passes clean_uploaded so we can skip the camera config lookup
+    # and the mediafiles import entirely on the common path (feature disabled)
+    if not clean_uploaded:
+        return
+
+    # upload succeeded and the camera removes its local copy after upload
+    try:
+        # local imports to avoid the config/mediafiles <-> uploadservices
+        # circular imports (both modules import uploadservices at module level)
+        from motioneye import config, mediafiles
+
+        camera_config = config.get_camera(camera_id)
+        rel = filename.removeprefix(camera_config['target_dir'] + os.sep)
+        mediafiles.del_media_content(camera_config, rel, media_type)
+        logging.debug(f'removed local file "{filename}" after successful upload')
+
+    except Exception as e:
+        # never let a cleanup error affect the (already successful) upload
+        logging.error(
+            f'failed to remove local file "{filename}" after upload: {e}',
             exc_info=True,
         )
 
--- a/motioneye/uploadservices.py
+++ b/motioneye/uploadservices.py
@@ -1356,7 +1356,7 @@ def upload_media_file(
         from motioneye import config, mediafiles
 
         camera_config = config.get_camera(camera_id)
-        rel = filename.removeprefix(camera_config['target_dir'] + os.sep)
+        rel = utils.remove_prefix(filename, camera_config['target_dir'] + os.sep)
         mediafiles.del_media_content(camera_config, rel, media_type)
         logging.debug(f'removed local file "{filename}" after successful upload')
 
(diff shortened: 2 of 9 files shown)


motioneye/uploadservices.py
```diff
@@ -1334,6 +1336,27 @@ def upload_media_file(camera_id, camera_name, target_dir, service_name, filename
...
+        # local imports to avoid the config/mediafiles <-> uploadservices
+        # circular imports (both modules import uploadservices at module level)
+        from motioneye import config, mediafiles
+
+        camera_config = config.get_camera(camera_id)
+        if camera_config.get('@clean_uploaded'):
+            local_target_dir = camera_config['target_dir']
+            rel = filename.removeprefix(local_target_dir + os.sep)
+            mediafiles.del_media_content(camera_config, rel, media_type)
+            logging.debug(f'removed local file "{filename}" after successful upload')
```
MichaIng: Since this is quite significant overhead, parsing the camera config here once again, I suggest to do this in the handler, which is the only point where `upload_media_file` is called. It might not look that clean, and requires passing more optional parameters, but should perform a lot better:
- Obtain `@clean_uploaded` in the `upload_media_file()` of the relay handler, and pass it as new `clean_uploaded: bool = False` to the `upload_media_file()` of the uploadservices.
- There it can be checked as cheap conditional before importing `mediafiles` and calling `del_media_content()`.

MichaIng (changes requested):
Looks good, see my only suggested change belog.

I wonder whether there is a suitable way to change the behavior of the cleanCloudEnabled switch, so that it does not take the existence of the local file as indicator whether to remove a file from the cloud, but to receive a file list with timestamps from the cloud, and do the cleanup independently. But that would require a new directory listing function for the UploadService class, and hence a sane API endpoint for each service. Yeah, definitely nothing for this PR, but it would be actually nice to keep the cloud cleanup after X days functional while files are locally removed immediately.

motioneye/uploadservices.py
JamBalaya56562: Good call — done in 41154b0.

`@clean_uploaded` is now read in the relay handler (which already has the camera config) and passed as `clean_uploaded: bool = False` to `uploadservices.upload_media_file`. The common path (feature disabled) returns immediately, **before** the `config.get_camera()` lookup and the `mediafiles` import.

`config.get_camera()` (which re-reads/parses the config file in the upload worker process, where the in-process cache is cold) now only runs when the feature is enabled and a file was actually uploaded — and there it's negligible next to the media-file read + the upload itself. I kept it in that branch because `del_media_content()` needs the camera config for `target_dir`; happy to also pass `target_dir` through if you'd prefer to drop the lookup entirely.

The tests now assert `get_camera` isn't even called on the disabled path.

JamBalaya56562:
Pushed the requested change (41154b0): `clean_uploaded` is now passed from the relay handler, so the disabled/common path no longer touches the camera config or imports `mediafiles`.

Re. the cloud-cleanup-by-timestamp idea — agreed it would be nice to keep "clean cloud after X days" working alongside immediate local removal. As you note, that needs a per-service remote directory-listing API on `UploadService`, so it's clearly a separate change; happy to open a follow-up issue to track it if useful.

Full test suite is green (121 passed). Ready for another look 🙂

motioneye/uploadservices.py
```diff
@@ -1334,6 +1342,30 @@ def upload_media_file(camera_id, camera_name, target_dir, service_name, filename
...
+
+    # upload succeeded and the camera removes its local copy after upload
+    try:
+        # local imports to avoid the config/mediafiles <-> uploadservices
+        # circular imports (both modules import uploadservices at module level)
+        from motioneye import config, mediafiles
+
+        camera_config = config.get_camera(camera_id)
+        rel = filename.removeprefix(camera_config['target_dir'] + os.sep)
+        mediafiles.del_media_content(camera_config, rel, media_type)
```
MichaIng, line 1360: Sadly `del_media_content` requires the camera config as input, so we still need to import `config`. But at least we do not need to obtain `target_dir`, since this is passed already:
```suggestion
        rel = filename.removeprefix(target_dir + os.sep)
        mediafiles.del_media_content(config.get_camera(camera_id), rel, media_type)
```
I also see `filename` is expected to be the full path, and validated to be within the target/camera dir: https://github.com/motioneye-project/motioneye/blob/dev/motioneye/handlers/relay_event.py#L73-L79
So obtaining the correct relative path the same way here is assured to not fail. It either fails validation in the handler, or deletion if the file does not exist at all.

JamBalaya56562: Thanks for the review! I looked into this and I think we need to keep `camera_config['target_dir']` here rather than the `target_dir` parameter.

The `target_dir` argument forwarded to `upload_media_file` is not the camera's `target_dir` — `relay_event.py` passes `camera_config['@upload_subfolders'] and camera_config['target_dir']` (L128-129). When a user disables `@upload_subfolders`, that short-circuits to `False`, which `upload_file` uses as the "no subfolders" signal (`if target_dir:` at L62).

So `filename.removeprefix(target_dir + os.sep)` would become `False + os.sep` → `TypeError` for anyone with subfolders disabled, which the cleanup `try/except` would swallow — silently breaking `@clean_uploaded` for that config (and the existing tests wouldn't catch it, since they always pass a real `target_dir`). Using `camera_config['target_dir']` keeps it correct regardless of the subfolders setting, and matches the prefix used by the validation at L75-79.

You're right that `del_media_content` only needs `target_dir`, but since its current signature takes the full config, I'll keep the `config.get_camera(camera_id)` lookup.

MichaIng: Oh you are absolutely right. We really need to add consequent typing across our code. It could check whether the value is None before obtaining it, but that does add more hassle than saving processing, given that the config is needed in any case.

MichaIng (approved):
LGTM

motioneye/uploadservices.py
```diff
@@ -1334,6 +1342,30 @@ def upload_media_file(camera_id, camera_name, target_dir, service_name, filename
...
+        return
+
+    # upload succeeded and the camera removes its local copy after upload
+    try:
+        # local imports to avoid the config/mediafiles <-> uploadservices
+        # circular imports (both modules import uploadservices at module level)
+        from motioneye import config, mediafiles
+
+        camera_config = config.get_camera(camera_id)
+        rel = filename.removeprefix(camera_config['target_dir'] + os.sep)
```
Marijn0: `str.removeprefix()` is only supported since Python 3.9:
https://docs.python.org/3/library/stdtypes.html#str.removeprefix

motionEye still declares support for Python >=3.7

motioneye/static/js/main.js
```diff
@@ -695,10 +695,20 @@ function initUI() {
...
         }
     });
+    $('#cleanUploadedSwitch').on('change', function () {
+        if (this.checked && $('#cleanCloudEnabledSwitch')[0].checked) {
+            /* mutually exclusive with cloud cleanup, which would delete the
+             * just-uploaded copies once the local files are gone */
+            $('#cleanCloudEnabledSwitch')[0].checked = false;
+            runAlertDialog(i18n.gettext('"Purigi la nubon" estis malŝaltita ĉar ĝi konfliktas kun "Forigi Dosierojn Post Alŝuto".'));
+        }
+    });
```
Marijn0: These two options are mutually exclusive, but the GUI button is not updated, so one switch does not visibly turn off the other.

motioneye/uploadservices.py
MichaIng: Ohh, good catch. We use it in the relay event as well.

I need to find a way to run tests with Python 3.7 in our CI. Since building (the wheel) requires Python 3.9 already, this is a bit difficult. `pre-commit` however seems to not complain.

```suggestion
        rel = filename[len(camera_config['target_dir']) :].lstrip(os.sep)
```
We do the same unconditional strip in `upload_file`, so that should be fine. Or you go the full failsafe way like here: https://github.com/motioneye-project/motioneye/blob/dev/motioneye/uploadservices.py#L62-L65

Please sync this with the relay handler.

motioneye/static/js/main.js
MichaIng: `$('#cleanCloudEnabledSwitch')[0].checked = false;` does not visibly toggle the switch?

... maybe `updateConfigUI()` is needed for this? But it looks more like controlling entirely hidden sections. Actually the `checked` attribute is supposed to work independently via CSS rules.

Marijn0: Not in my testing, AFAICS. Setting `.checked = false` does not visibly update the styled switch; it needs to trigger the checkbox change handler, something like this:

`$('#cleanCloudEnabledSwitch').prop('checked', false).trigger('change');`
`$('#cleanUploadedSwitch').prop('checked', false).trigger('change');`

Marijn0:
I was only able to test it with an FTP server, but it works quite nicely. Many thanks, @JamBalaya56562!

motioneye/uploadservices.py
JamBalaya56562: Good catch on the 3.7 incompatibility — fixed.

One subtlety though: a plain `filename[len(...):]` slice would break the absolute-path rejection in `validate_paths`. `str.removeprefix` returns the string **unchanged** when it doesn't start with the prefix, and that is exactly what lets a foreign/absolute path (e.g. `/etc/passwd`, which is not under `target_dir`) reach `validate_paths` still looking absolute and get rejected. A blind slice strips `len(target_dir)` chars off *any* path, turning `/etc/passwd` into a relative-looking string that sneaks past the check — `tests/test_mediafiles.py::TestMediaFilesPathValidation::test_make_movie_preview_rejects_absolute_path` fails with it.

So I added a faithful backport `utils.remove_prefix()` (unchanged on no-match) and used it in `uploadservices`, the relay handler (synced as requested), and `mediafiles.make_movie_preview`, which had the same `str.removeprefix` usage. Added `tests/test_utils/test_remove_prefix.py` to lock in the no-match behaviour. (a7cf1ab)

motioneye/static/js/main.js
JamBalaya56562: You're right, confirmed — setting `.checked = false` alone doesn't repaint the styled switch. Switched both mutual-exclusion handlers to `.prop('checked', false).trigger('change')` as you suggested (a7cf1ab).

The guards (`if (enabled)` / `if (this.checked && ...)`) evaluate against the now-false state, so triggering the other switch's `change` doesn't recurse. Thanks for testing this!

Marijn0 (approved):

motioneye/uploadservices.py
MichaIng: Probably best to emulate `removeprefix()`. However, it is not possible to bypass the path validation. That checks explicitly whether the passed path is inside the target/camera dir. So if an invalid path is passed, it will either still fail the validation, or, if the blind slicing does result in a relative path without traversal elements, effectively point to a legal location inside the camera dir, resulting in a failure as the path will (most likely) just not exist. The conditional splicing hence results in a more meaningful error message.

> `tests/test_mediafiles.py::TestMediaFilesPathValidation::test_make_movie_preview_rejects_absolute_path` fails with it.

For the piggyback removeprefix case in the relay event you mean, right, that is probably the stronger reason to pass through invalid paths unmodified.

_I would actually prefer if the API itself would take a relative path. Every API endpoint should always accept paths relative to the camera dir only. There is no action involving a path that is not related to a particular camera, at least those relay actions always are. That would make it a lot easier, rendering all the path forging and slicing obsolete, allowing validation in any stage or every stage. However, not something to change in this PR.

... and I guess the reason for the relay event to take an absolute path is that `motion` calls the API in this case, which does not know about motionEye camera dirs, and just passes its `%f` conversion specifier. We'd need to split off the camera dir in https://github.com/motioneye-project/motioneye/blob/dev/motioneye/scripts/relayevent.sh._

MichaIng (approved):

PR #3376, opened 2026-06-28
fix(dropbox): encode test_access JSON body as bytes

### LINKED ISSUE #2828: Dropbox - Service error
Hello,
i'm configuring dropbox backup and I'm getting after the key handshaking the following error

ERROR: dropbox: request failed: POST data should be bytes, an iterable of bytes, or a file object. It cannot be of type str.

### LINKED ISSUE #2940: Motioneye Google Drive - Dropbox connection
Just installed Motioneye and see it’s no longer possible to connect to Google Drive or Dropbox.

Does anybody know if it will be resolved, all info after searching seems to point that it’s dead.

Any news?
Thanks

JamBalaya56562:
## What

`Dropbox.test_access()` sent a **str** JSON body to `urllib`'s `Request`, which rejects it with:

```
POST data should be bytes, an iterable of bytes, or a file object. It cannot be of type str.
```

So testing/configuring the Dropbox upload service fails right after the OAuth handshake. This encodes the JSON body to `bytes`.

## Why

`uploadservices.py` `Dropbox.test_access()` built the list-folder request body with `json.dumps(body)` (a `str`) and passed it to `self._request()` → `Request(url, data=body)`. `urllib` requires `data` to be bytes. The credential paths (`_request_credentials` / `_refresh_credentials`) already `.encode()` their bodies, and `upload_data()` passes the raw binary file, so `test_access` was the only str-body path.

Fixes #2828 — confirmed by multiple reporters (the same `.encode()` fix was suggested in the thread).

## How

```python
-        body = json.dumps(body)
+        body = json.dumps(body).encode()
```

## Testing

- New `tests/test_uploadservices_dropbox.py` asserts `test_access()` sends a `bytes` body to `_request` (and that it round-trips to the expected JSON).
- Full suite green: **117 passed** (Linux). `ruff check` / `ruff format --check` pass.

## Notes / out of scope

- A reporter also noted the test-result UI then shows `Accessing the upload service succeeded![object Object]` — a separate cosmetic JS message-handling issue, not addressed here.
- `GooglePhoto` builds some request bodies with `json.dumps(...)` without `.encode()` as well; left untouched since it's a different service (and Google OOB OAuth has its own issues), but it may warrant a follow-up.

Closes #2828

### DIFF
--- a/motioneye/uploadservices.py
+++ b/motioneye/uploadservices.py
@@ -705,7 +705,9 @@ class Dropbox(UploadService):
             'include_deleted': False,
         }
 
-        body = json.dumps(body)
+        # the body must be bytes: urllib's Request rejects a str POST body
+        # with "POST data should be bytes ... not str" (#2828)
+        body = json.dumps(body).encode()
         headers = {'Content-Type': 'application/json'}
 
         try:
--- a/tests/test_uploadservices_dropbox.py
+++ b/tests/test_uploadservices_dropbox.py
@@ -1,0 +1,44 @@
+# Copyright (c) 2013 Calin Crisan
+# This file is part of motionEye.
+#
+# motionEye is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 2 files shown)


MichaIng (approved):
Right, the same fix was already done for Google uploads a while ago: https://github.com/motioneye-project/motioneye/commit/9ad71b2
AFAIK this is still a Python 2 thing, where string data was accepted. And obviously it was never fixed.

PR #3377, opened 2026-06-29
Allow 12-hour/AM-PM strftime specifiers (%I, %p, %l) in file names

### LINKED ISSUE #2070: Is there a filepath/filename length limit in the UI?
I want to use the following for "image file name":
camname_%Y-%m-%d/camname_%Y-%m_%d_%I-%M-%S-%p_%q

and this for "movie file name":
camname_%Y-%m-%d/camname_%Y-%m_%d_%I-%M-%S-%p

but upon clicking the apply button, motioneye throws up the message
"Make sure all the configuration options are valid!"

When I look at the fields, they have the little red warning icon / mouseover text "special characters are not allowed in the filename", but I'm pretty sure I'm not using any because:
a) if I delete some from anywhere, eventually the warning vanishes
b) if I manually populate the filename in camera-1.conf, it works fine. files are saved with the right names in the right place.

Might be related to https://github.com/ccrisan/motioneye/issues/1497 but I dunno.

Thanks,
Matt

JamBalaya56562:
## What

The image/movie **file name** validator rejected the `%I` (12-hour hour), `%p` (AM/PM) and `%l` (space-padded 12-hour hour) Motion conversion specifiers. Patterns such as `%Y-%m-%d/%I-%M-%S-%p` were refused in the UI with *"Make sure all the configuration options are valid!"*, even though Motion saves the files correctly when the same pattern is set directly in the config.

## Why

`filenameValidRegExp` only whitelisted `%[CYmdHMSqv$]`. `%I`, `%p` and `%l` are valid `strftime` / Motion conversion specifiers that simply weren't in the allow-list, as confirmed in the issue discussion.

## How

- Add `I`, `p`, `l` to `filenameValidRegExp` in `motioneye/config.py` **and** `motioneye/static/js/main.js` (kept in sync, per the existing in-code requirement that the two regexes match).
- List the new specifiers in the validation failure message.

## Testing

Verified the updated regex (both Python and JS) :
- accepts the reported patterns, e.g. `camname_%Y-%m-%d/camname_%Y-%m_%d_%I-%M-%S-%p_%q` and `%l-%M-%S`
- still accepts existing patterns, e.g. `%Y-%m-%d/%H-%M-%S-%q`
- still rejects invalid input, e.g. `%Z`, `%y`, `;`

Fixes #2070

🤖 Generated with [Claude Code](https://claude.com/claude-code)


MichaIng:
I do not see any of those documented at https://motion-project.github.io/motion_config.html#conversion_specifiers 🤔.

EDIT: Ah sorry, part of `strftime`:

> In addition to the above, the conversion specifiers include the same options as for the C function strftime (3).

MichaIng (changes requested):
This is generally fine, but we should update all translation files accordingly, so translators do not need to re-translate the whole string. Check how it was done here: https://github.com/motioneye-project/motioneye/pull/3273/changes

JamBalaya56562:
Done in 24cdeead — updated all `motioneye.po` catalogs (32) plus `motioneye.pot`, inserting `%I %p %l` into the `filenameFailMessage` specifier list in both the `msgid` and the translated `msgstr` (where one exists), exactly like #3273. Files with an empty `msgstr` only got the `msgid` updated.

Since the specifiers are literal, the existing translations stay valid and translators don't need to redo the string. I confirmed the updated `msgid` still matches the source string in `config.py`, so gettext keeps resolving the translations.

MichaIng (approved):
Great, all fine now!

PR #3384, opened 2026-06-30
fix(media): use pre-capture offset for movie previews

### LINKED ISSUE #2766: Thumbnail offset hardcoded weirdly?
Hi!

Ive been having issues that Motioneye generates thumbnails that are not representative of the actual detection event. I checked the code and found out theres a a minimum offset, and also a weirdly doubled offset for precapture?

For example if i have a camera with framerate of 10, and precapture of 40 frames before a detection, we end up with the thumbnail being generated 8 seconds from the video, eventhough the detection is after the 40 precapture frames, meaning the thumbnails 4 seconds late.

I managed to get alot better thumbnails when i changed the offs value to just be offs +=1 so the thumbnails are taken 1 second after detection.

This seems to be done deliberately so maybe im missing the point of this? is there any reason for it to do this? 

https://github.com/motioneye-project/motioneye/blob/03882a18994fc35c22ae169dd1569abb923c7cef/motioneye/mediafiles.py#L349-L363

JamBalaya56562:
## Summary
- use the configured pre-capture offset directly when creating movie preview thumbnails
- remove the empirical doubled-offset/duration selection logic
- add a unit test for the `pre_capture / framerate` thumbnail offset

## Why
The old logic preferred `pre_capture / framerate * 2` when it fit within the video duration, which could place thumbnails several seconds after the actual motion event. The configured pre-capture value already identifies where the event begins in the recorded movie.

Fixes #2766.

## Validation
- `python3 -m py_compile motioneye/mediafiles.py tests/test_mediafiles.py`
- Could not run the unittest target locally because the environment is missing `tornado`, imported by the test package

### DIFF
--- a/motioneye/mediafiles.py
+++ b/motioneye/mediafiles.py
@@ -428,36 +428,10 @@ def cleanup_media(media_type: str) -> None:
         )
 
 
-def get_movie_duration_seconds(path: str) -> int:
-    cmd = [
-        'ffprobe',
-        '-v',
-        'error',
-        '-show_entries',
-        'format=duration',
-        '-of',
-        'default=noprint_wrappers=1:nokey=1',
-        path,
-    ]
-
-    try:
-        result = utils.call_subprocess(cmd, stderr=None)
-        duration_seconds = int(float(result))
-        return duration_seconds
-
-    except Exception as e:
-        logging.error(f'failed to determine duration of {path}: {e}')
-        return 0
-
-
 def make_movie_preview(camera_config: dict, full_path: str) -> Optional[str]:
     framerate = camera_config['framerate']
     pre_capture = camera_config['pre_capture']
     offs = pre_capture / framerate
-    movie_duration = get_movie_duration_seconds(full_path)
-    offs = max(
-        (x for x in {offs * 2, offs} if x <= movie_duration), default=movie_duration / 2
-    )
 
     path = quote(full_path)
     thumb_path = full_path + '.thumb'
--- a/tests/test_mediafiles.py
+++ b/tests/test_mediafiles.py
@@ -20,6 +20,7 @@ from pathlib import Path
 from shutil import rmtree
 from tempfile import mkdtemp
 from time import time
+from unittest.mock import patch
 
 from motioneye import mediafiles
 from motioneye.mediafiles import _list_media_files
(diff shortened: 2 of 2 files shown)


zagrim:
I wrote a comment on the related issue. Basically, I don't think a good solution cannot be this simple, and this change might break things for some users for which the earlier behaviour has been mostly ok.

MichaIng (changes requested):
I guess the reason for the 2x pre-capture (2nd frame with motion) offset is to not use the very first frame with motion, where the moving object might be just visible with a tiny part. There is a good example in the referenced issue: https://github.com/motioneye-project/motioneye/issues/2766#issuecomment-1558361467

Since 2x pre-capture was the fixed only offset earlier, which defaults to 4s, failing for shorter movies, we did add the movie duration check, the additional 1x pre-capture (first frame with motion), and movie duration / 2 as fallback value quite recently, so thumbnail generation does not fail for short movies.

Whether using the 1st frame with motion or the 2nd by default can be discussed, but we need to keep the movie length measurement in any case: The pre-capture duration can be changed, hence it is perfectly possible that a movie is shorter than the pre-capture time. Also other short videos could be copied into the data dir. We should assure thumbnails can successfully generate in any case.

Generally, I think the current behaviour is fine, and if we change something, we should let users decide up to how many frames offset into the movie they want to use for thumbnails

JamBalaya56562:
Thanks @MichaIng and @zagrim for the detailed reviews — you're both right, and I'm closing this.

Two things I got wrong:

- **Removing the duration measurement was a regression.** I read the `2x → 1x → duration/2` chain as leftover empirical logic, but it is the recent v0.44.0 fix for movies shorter than the pre-capture time — which, as you point out, can happen whenever `pre_capture` is changed, or when another short video lands in the data dir. It needs to stay.
- **The approach lands on the wrong frame anyway.** Re-reading #2766, @suvis1 asked for detection + ~1s and demonstrated with images that the detection moment itself only shows a sliver of the subject. `offs = pre_capture / framerate` puts the thumbnail exactly on that frame, so this PR would not even solve the original report.

On the way forward: you both converged on making the offset user-selectable, with the open question being its form — @zagrim's 2–3 preset modes vs @MichaIng's freely selectable frames/seconds. I'll leave that design call to you in #2766 rather than guess at it. If you settle on a direction there, I'm happy to pick up the implementation, keeping `get_movie_duration_seconds()` and the short-movie fallback intact as the safety net.

PR #3386, opened 2026-06-30
fix(config): add "pause" to used motion options

MichaIng:
It was added for motion >= 4.4 as replacement for the API call, to disable motion detection. Options not listed in _USED_MOTION_OPTIONS are however shown in the extra options box of the GUI, which can be confusing. Make it a known option, as it will be always set explicitly if supported by the motion version.

### DIFF
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -107,6 +107,7 @@ _USED_MOTION_OPTIONS = {
     'on_event_start',
     'on_movie_end',
     'on_picture_save',
+    'pause',
     'picture_filename',
     'picture_output_motion',
     'picture_output',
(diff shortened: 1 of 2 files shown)


MichaIng:
@Marijn0 FYI, as recognized in #3372. It has no effect on the position within the config file, which is defined by the place at which the key is declared, i.e. where (and if) it is in `_set_default_motion_camera()`. But since it is not supported by motion < 4.4, and actively removing it before writing the config file might cause more confusion than that it helps, we shouldn't set a default for now, but keep `pause` just an inverse of `@motion_detection`, only set if motion >= 4.4. Most importantly, it does not show up in the extra options box, where it is confusing, indeed.

PR #3387, opened 2026-06-30
fix(ui): hide disabled media browser buttons when empty

### LINKED ISSUE #2731: Disable/hide button "open pictures browser" possible

JamBalaya56562:
## What

Hide a media-browser button (Pictures / Movies) in the camera overlay when its capture mode is disabled **and** there are no existing files of that type to browse.

Fixes #2731.

> Re-created after the previous PR (#3383) was accidentally closed when its branch was deleted during a rebase. This branch carries the change @Marijn0 requested in that review.

## Why

The overlay showed the media-browser buttons even when the corresponding capture feature was disabled, which is inconsistent with the camera settings. Simply hiding on the config flag (the first approach) would also cut off access to media recorded *before* the user disabled capture — per @Marijn0's review, the button should disappear only when it truly has nothing to show.

## How

- Simple MJPEG cameras still hide both buttons (unchanged).
- Otherwise, when a capture mode is disabled, an async check hits the existing list endpoint (`…/list/?with_stat=false`) and the button is hidden **only if `mediaList` comes back empty**. Enabled modes issue no request; on error / while unsure the button stays visible (no flicker for the case that still has media).

## Validation

- Change is limited to `motioneye/static/js/main.js` (+22/-0); `node --check` passes.
- No JS test framework in the repo. Manual E2E (needs a live motion+camera): disable Still images on a camera with old pictures → Pictures button stays; with none → it hides after load; same for Movies; simple MJPEG still hides both.

## Screenshots

Setup used for the comparison: one JPEG network camera, **Still Images and Movies both disabled**, and **no media files** stored for the camera. (Same camera/config on both servers; only the motionEye code differs.)

| Before (current `dev`) | After (this PR) |
|---|---|
| before | after |
| The Pictures and Movies buttons are shown in the camera overlay even though both capture modes are disabled and there is nothing to browse. | The Pictures and Movies buttons are hidden, because the `…/list/?with_stat=false&limit=1` check found no existing media. The other overlay buttons are unaffected. |

### DIFF
--- a/motioneye/static/js/main.js
+++ b/motioneye/static/js/main.js
@@ -4835,6 +4835,12 @@ function addCameraFrameUi(cameraConfig) {
         picturesButton.hide();
         moviesButton.hide();
     }
+    else if (!cameraConfig['still_images']) {
+        picturesButton.hide();
+    }
+    if (cameraConfig['proto'] != 'mjpeg' && !cameraConfig['movies']) {
+        moviesButton.hide();
+    }
 
     cameraFrameDiv.attr('id', 'camera' + cameraId);
     cameraFrameDiv[0].refreshDivider = 0;
--- a/motioneye/static/js/main.js
+++ b/motioneye/static/js/main.js
@@ -4721,9 +4721,11 @@ function runMediaDialog(cameraId, mediaType) {
 
 function hideMediaButtonIfEmpty(cameraId, mediaType, button) {
     /* the list endpoint returns {mediaList: [...]} (empty when there are no
-     * files); with_stat=false keeps the request cheap. On error the button is
-     * left visible (safe default - don't hide when we are unsure). */
-    ajax('GET', basePath + mediaType + '/' + cameraId + '/list/?with_stat=false', null, function (data) {
+     * files); limit=1 makes the server stop walking the media directory at the
+     * first file found, so this stays cheap even with many stored files. On
+     * error the button is left visible (safe default - don't hide when we are
+     * unsure). */
+    ajax('GET', basePath + mediaType + '/' + cameraId + '/list/?with_stat=false&limit=1', null, function (data) {
         if (data && data.mediaList && !data.mediaList.length) {
             button.hide();
         }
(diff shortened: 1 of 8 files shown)


motioneye/static/js/main.js
```diff
@@ -4719,6 +4719,17 @@ function runMediaDialog(cameraId, mediaType) {
...
+function hideMediaButtonIfEmpty(cameraId, mediaType, button) {
+    /* the list endpoint returns {mediaList: [...]} (empty when there are no
+     * files); with_stat=false keeps the request cheap. On error the button is
+     * left visible (safe default - don't hide when we are unsure). */
+    ajax('GET', basePath + mediaType + '/' + cameraId + '/list/?with_stat=false', null, function (data) {
+        if (data && data.mediaList && !data.mediaList.length) {
+            button.hide();
+        }
+    });
+}
```
Marijn0, line 4733: I’m not sure we should call the media list endpoint just to decide whether to hide a button. Even with `with_stat=false`, this can still be resource-intensive for users with lots of saved files. Unfortunately, I don’t see an easy alternative solution either.

JamBalaya56562: Thanks, agreed that calling the full media list just to decide whether to hide the button would be too expensive for large media directories.

I addressed this in [`c23c3d8`](https://github.com/motioneye-project/motioneye/pull/3387/commits/c23c3d8ad9169ffd5c2d8f0ba2695da298bba864) by turning that request into a bounded existence check instead of a full listing:

- the UI now calls `/list/?with_stat=false&limit=1` for the hide-button check;
- the picture/movie list handlers parse an optional positive `limit` query parameter;
- local media listing stops walking as soon as the requested number of matching files is found, so `limit=1` returns after the first actual media file;
- the limit is also respected across recursive subdirectories;
- remote camera list requests forward `with_stat=false` and `limit=1` only when explicitly requested, while the default request remains unchanged for compatibility with older remote motionEye instances.

I also added tests for the early-stop behavior, handler query parsing, recursive limit handling, and remote query forwarding.

Tested with WSL/Linux:

```text
pytest tests/test_mediafiles.py tests/test_remote.py tests/test_handlers/test_media_list_limit.py -q
65 passed, 123 subtests passed

pytest tests -q
146 passed, 10 warnings, 123 subtests passed
```

PR #3388, opened 2026-07-01
enh(mediafiles): eliminate shell=True subprocess calls

MichaIng:
* For the ffmpeg codecs list with hardcoded command, it was safe but unnecessary.
* When defining the command as list in the first place, `shlex.quote()` is not needed, and `.split()` can be skipped.
* For timelapse movies from a list of pictures, the FFmpeg's concat extension can be used instead of image2pipe, which allows to pipe the list of files to ffmpeg from Python, rather than the content of the files from shell, so that `shell=True` is not needed. Adding `-y` to let ffmpeg overwrite an existing output file automatically, renders the prior file removal obsolete as well.

This commit adds some typing as well.

### DIFF
--- a/motioneye/mediafiles.py
+++ b/motioneye/mediafiles.py
@@ -25,10 +25,9 @@ import subprocess
 from errno import EAGAIN, ENOENT
 from hashlib import sha1
 from io import BytesIO
-from shlex import quote
 from signal import SIGKILL, SIGTERM
 from time import time
-from typing import Awaitable, List, Optional
+from typing import Any, Awaitable, List, Optional
 from zipfile import ZipFile
 
 from PIL import Image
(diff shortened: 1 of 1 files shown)


zagrim (approved):
👍

MichaIng:
Good to know: `subprocess.run()` (hence our `utils.call_subprocess()`) allows strings or path-like objects as command arguments only, when passing the command as list. `utils.call_subprocess(['echo', 'We count:', 1, 2, 3, '...')` hence does not work. It needs to be `'1', '2', '3'`.

MichaIng:
Wouldn't hurt if someone else tests the timelapse feature as well, as it has gone quite through some changes.

When I tried to debug a timelapse generation failure, I recognized that we do not log `ffmpeg`'s error output at all. It required some structural change to collect the output from the STDOUT/STDERR pipe with each poll, but it works well now and enabled me to debug the actual `ffmpeg` error.

Apart of the complicated `concat` module safety options, the final error was simply a wrong HEVC codec name, which is `hevc`, not `h265`. Also for other codecs in that mapping, we actually use encoder/decoder names, which can be different. Hence could make sense to test with H.264, and, where possible, H.264/V4L2MPM. The latter fails on RPi 5, as they removed the H.264 hardware codec, but was a good encouragement to test with HEVC.

PR #3389, opened 2026-07-02
Filter config listing by user role

Marijn0:
- Add an explicit whitelist of camera fields for normal users.
- Apply the whitelist to local, remote, and simple MJPEG camera entries.
- Add tests covering normal, admin, and peer user types.

### DIFF
--- a/motioneye/handlers/config.py
+++ b/motioneye/handlers/config.py
@@ -45,6 +45,30 @@ __all__ = ('ConfigHandler',)
 
 
 class ConfigHandler(BaseHandler):
+    # fields visible to the normal/surveillance user
+    _NON_ADMIN_CAMERA_FIELDS = frozenset(
+        {
+            'id',
+            'name',
+            'enabled',
+            'proto',
+            'actions',
+            'streaming_framerate',
+            'streaming_server_resize',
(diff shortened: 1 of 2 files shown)


motioneye/handlers/config.py
```diff
@@ -440,18 +464,20 @@ def _handle_get_config_response(
...
+                        'enabled': False,
+                        'connection_failed': True,
+                        'connection_error': msg,
+                        'connection_url': remote.pretty_camera_url(
+                            local_config, camera=False
+                        ),
+                        'streaming_framerate': 1,
+                        'framerate': 1,
+                    }
+                )
```
MichaIng: This filters out the `connection_*` values, which are consumed by the frontend to handle and show remote motionEye camera connection errors, isn't it?

And since this is a limited list of fields we fully control, when the remote motionEye cam config could not be received, I think this filtering is not needed here?

Marijn0: Yeah, I can get behind that. I'll remove it. 👍

MichaIng (approved):
I did not test it yet, but looks pretty good and tests are sane.

zagrim (approved):
👍

PR #3391, opened 2026-07-06
fix for v4l2 cameras: show bool controls and fix control name mismatch with Motion

### LINKED ISSUE #1456: V4L2 video controls
Hi,

is that only with me or does the V4L2 setting with others not work?

"all V4L2 video controls can now be adjusted from the UI" since 0.40

Unfortunately, the controllers have no effect.

System:
Raspberry Pi 4
motionEye Version 0.41
Motion Version 4.1.1
OS version Raspbian 10
Pi Cam

thanks
Daniel

### LINKED ISSUE #2300: motioneye not applying v4l2 config settings
I was trying to adjust the exposure settings for the camera I am using.

when doing
```
v4l2-ctl -c exposure_auto=1 (for manual mode)
v4l2-ctl -c exposure_absolute=20
```

the picture looks perfectly fine. but when i try to set the same setting in the "Video Device" menu in motioneye is simply doesn't change anything.

After changing the setting in the webinterface `v4l2-ctl -L` does not reflect the changes which i assume it should if they were applied correctly, right?

The other settings also seem to be kinda buggy. sometimes they apply, sometimes they don't and I can't really see a pattern here.
Anyone one have any idea what is wrong here?

Marijn0:
`v4l2-ctl --list-ctrls` reports boolean controls, such as `white_balance_automatic`, without `min`, `max`, or `step` properties. Integer controls usually include these properties, as shown below.

`list_ctrls()` already matched the parenthesized control type, but did not capture or use it. Downstream, `config.py` filters out controls without min and max properties, while the frontend renders controls with `min == 0 && max == 1` as checkboxes instead of range sliders.

Capture the V4L2 control type and default bool controls to `min=0` and `max=1`, allowing them to pass through the existing filter and render as checkboxes (just like was already done for `int` control options with only two states, such as `backlight_compensation` below).

Example `v4l2-ctl --list-ctrls`:
```
User Controls

                     brightness 0x00980900 (int)    : min=-64 max=64 step=1 default=0 value=0
                       contrast 0x00980901 (int)    : min=0 max=100 step=1 default=50 value=50
                     saturation 0x00980902 (int)    : min=0 max=100 step=1 default=64 value=64
                            hue 0x00980903 (int)    : min=-180 max=180 step=1 default=0 value=0
        white_balance_automatic 0x0098090c (bool)   : default=1 value=1  <---
                          gamma 0x00980910 (int)    : min=100 max=500 step=1 default=300 value=300
           power_line_frequency 0x00980918 (menu)   : min=0 max=2 default=1 value=1 (50 Hz)
      white_balance_temperature 0x0098091a (int)    : min=2800 max=6500 step=10 default=4600 value=4600 flags=inactive
                      sharpness 0x0098091b (int)    : min=0 max=100 step=1 default=50 value=50
         backlight_compensation 0x0098091c (int)    : min=0 max=1 step=1 default=0 value=0

Camera Controls

                  auto_exposure 0x009a0901 (menu)   : min=0 max=3 default=3 value=3 (Aperture Priority Mode)
         exposure_time_absolute 0x009a0902 (int)    : min=50 max=10000 step=1 default=166 value=166 flags=inactive
     exposure_dynamic_framerate 0x009a0903 (bool)   : default=0 value=1  <---
```
#
Also, many controls never actually reached the camera. Motion matches `video_params` entries against the control's raw driver name (as returned by `VIDIOC_QUERYCTRL`), not the name `v4l2-ctl` shows (used by motionEye). For controls with a multi-word name, these two differ: the driver reports `"White Balance, Automatic"`, while `v4l2-ctl` normalizes that to `white_balance_automatic`. Motion's name comparison doesn't account for this, so it silently never applied the setting without an error, the value just never changed beyond the config.

Single-word controls (`brightness`, `contrast`, ...) happened to still work, since there's nothing to normalize, which is likely why this went unnoticed for so long.

Instead of the name, write the control's numeric ID, the same [ID%08d](https://github.com/Motion-Project/motion/blob/release-4.7.1/src/video_v4l2.c#L240) format Motion itself uses internally (`ctrl_iddesc` in `video_v4l2.c`) to identify a control regardless of its name (and is listed in the motion documentation: https://motion-project.github.io/motion_config.html#video_params). This matches unambiguously, whatever the driver calls the control. (and was just the easiest fix 😅)

Old (name-based, silently ignored by Motion for multi-word names):

    video_params white_balance_automatic=0,brightness=30,exposure_dynamic_framerate=1

New (ID-based, always matches):

    video_params ID09963788=0,ID09963776=30,ID10094851=1

### DIFF
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -1120,9 +1120,12 @@ def motion_camera_ui_to_dict(ui, prev_config=None):
         threshold = int(float(ui['frame_change_threshold']) * width * height / 100)
 
         if proto == 'v4l2':
-            # video controls
+            # video controls, control_id is the ID Motion actually matches against (see v4l2-ctl --list-ctrls)
+            video_controls = v4l2ctl.list_ctrls(prev_config.get('videodevice', ''))
             vid_control_params = (
-                ('{}={}'.format(n, c['value']))
+                '{}={}'.format(
+                    video_controls.get(n, {}).get('control_id', n), c['value']
+                )
                 for n, c in list(ui['video_controls'].items())
             )
             data['vid_control_params'] = ','.join(vid_control_params)
@@ -1682,6 +1685,12 @@ def motion_camera_dict_to_ui(data):  # noqa: C901
         ui['resolution'] = str(data['width']) + 'x' + str(data['height'])
 
         video_controls = v4l2ctl.list_ctrls(data['videodevice'])
+
+        # video_params is keyed by control_id, map it back to our UI name
+        name_by_control_id = {
+            c.get('control_id', n): n for n, c in video_controls.items()
+        }
+
         video_controls = [
             (n, c)
             for (n, c) in list(video_controls.items())
@@ -1701,6 +1710,7 @@ def motion_camera_dict_to_ui(data):  # noqa: C901
             else:
                 continue  # ignore any other kind of param
 
+            name = name_by_control_id.get(name, name)
             vid_control_values[name] = value
 
         ui['video_controls'] = {
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -1122,13 +1122,14 @@ def motion_camera_ui_to_dict(ui, prev_config=None):
         if proto == 'v4l2':
             # video controls, control_id is the ID Motion actually matches against (see v4l2-ctl --list-ctrls)
             video_controls = v4l2ctl.list_ctrls(prev_config.get('videodevice', ''))
-            vid_control_params = (
-                '{}={}'.format(
-                    video_controls.get(n, {}).get('control_id', n), c['value']
+            if video_controls:
+                vid_control_params = (
+                    '{}={}'.format(
+                        video_controls.get(n, {}).get('control_id', n), c['value']
+                    )
+                    for n, c in list(ui['video_controls'].items())
                 )
-                for n, c in list(ui['video_controls'].items())
-            )
-            data['vid_control_params'] = ','.join(vid_control_params)
+                data['vid_control_params'] = ','.join(vid_control_params)
 
     else:  # assuming netcam
         if match(
--- a/motioneye/controls/v4l2ctl.py
+++ b/motioneye/controls/v4l2ctl.py
@@ -195,7 +195,13 @@ def list_ctrls(device):
         # Motion matches controls by ID. We use ID because the driver's raw
         # name doesn't always match the name v4l2-ctl gives (e.g. driver:
         # "White Balance, Automatic" vs. v4l2-ctl: "white_balance_automatic")
-        properties['control_id'] = f'ID{int(id_hex, 16):08d}'
+        if id_hex:
+            try:
+                properties['control_id'] = f'ID{int(id_hex, 16):08d}'
+            except ValueError:
+                logging.warning(
+                    f'could not parse v4l2 control id "{id_hex}" for "{control}"'
+                )
 
         controls[control] = properties
 
--- a/motioneye/controls/v4l2ctl.py
+++ b/motioneye/controls/v4l2ctl.py
@@ -179,14 +179,18 @@ def list_ctrls(device):
         if not line:
             continue
 
-        match = re.match(r'^\s*(\w+)\s+([a-f0-9x\s]+)?\(\w+\)\s*:\s*(.+)\s*', line)
+        match = re.match(r'^\s*(\w+)\s+([a-f0-9x\s]+)?\((\w+)\)\s*:\s*(.+)\s*', line)
         if not match:
             continue
 
-        control, _, properties = match.groups()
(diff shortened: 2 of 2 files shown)


Marijn0:
Interesting, a lot of v4l2 control options values don't actually reach the camera, because of a name mismatch between how motionEye writes them to the config and how Motion reads them back.

A bit unrelated to this fix, just ran into it while testing. Still need to dig further into the exact root cause, but it looks like it might be a quirk on Motion's side. 🤔

Marijn0:
So I found the cause for the camera options not applying: it was indeed a name mismatch between motionEye and Motion, which was inherent in the way the control options were obtained. I switched to writing the options based on their control ID, which is supported by Motion, as can be read in the documentation: 
https://motion-project.github.io/motion_config.html#video_params

The only slight downside is that the config options are harder to read for a user editing the config manually. Not sure if this is common practice, or even recommended, when you don't know the min/max or supported values for a given option.

motioneye/config.py
```diff
@@ -1123,9 +1123,12 @@ def motion_camera_ui_to_dict(ui, prev_config=None):
...
+            # video controls, control_id is the ID Motion actually matches against (see v4l2-ctl --list-ctrls)
+            video_controls = v4l2ctl.list_ctrls(prev_config.get('videodevice', ''))
             vid_control_params = (
-                ('{}={}'.format(n, c['value']))
+                '{}={}'.format(
+                    video_controls.get(n, {}).get('control_id', n), c['value']
+                )
                 for n, c in list(ui['video_controls'].items())
             )
             data['vid_control_params'] = ','.join(vid_control_params)
```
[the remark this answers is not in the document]
Marijn0: In `motion_camera_ui_to_dict()`, the UI control name is looked up via `list_ctrls()` so the corresponding V4L2 control ID can be written to the config in the format expected by Motion.

If that lookup fails because the camera connection is lost between loading the settings page and saving it, the UI control name would be written to the config instead, so I think it makes sense to avoid that by leaving the current `vid_control_params` untouched when the control list cannot be read, especially since Motion would probably not be able to apply updated control values reliably at that point anyway.

I'm not sure whether it would be better to pass the ID together with the name from `motion_camera_dict_to_ui()` and pass it back into `motion_camera_ui_to_dict()` to avoid a second lookup when saving.

Edit: I think it's fine for now.

motioneye/config.py
```diff
@@ -1120,12 +1120,16 @@ def motion_camera_ui_to_dict(ui, prev_config=None):
...
         if proto == 'v4l2':
-            # video controls
-            vid_control_params = (
-                ('{}={}'.format(n, c['value']))
-                for n, c in list(ui['video_controls'].items())
-            )
-            data['vid_control_params'] = ','.join(vid_control_params)
+            # video controls, control_id is the ID Motion actually matches against (see v4l2-ctl --list-ctrls)
+            video_controls = v4l2ctl.list_ctrls(prev_config.get('videodevice', ''))
+            if video_controls:
```
MichaIng, line 1125: @Marijn0 
Great work with the back and forth mapping. Regarding this condition here, I see this was discussed with Copilot already:
- `videodevice` (`video_device` is modern motion config) is moreless assured to be set, right? Mandatory to fetch the stream, and to apply controls.
- `v4l2ctl.list_ctrls()` can fail if the camera is not accessible, so control IDs cannot be retrieved. So parameters would be overwritten with the names again. So it makes sense to prevent that.

But, we should probably not do this silently, but throw a warning, or maybe even a UI prompt? Something to indicate that the camera needs to be accessible to store/change controls, and that access failed. AFAICS, it would currently log something like `failed to list controls of device "/dev/video0"` already, but not directly indicating, that hence the controls from UI couldn't be applied. Maybe below adding the following:
```py
else:
    logging.warning(f'could not apply controls for camera id "{prev_config['@id']}"')
```
or something like that? Frontend prompt is probably more complicated to achieve from here.

Marijn0: Sure, I'll open a PR adding it, and indeed an additional frontend warning is not that straightforward from here AFAICS. I'm also not sure the control listing is expected to fail while the video stream is available, that is the case where such a warning in the frontend would really add value. A camera that is not available would be a bigger concern than some config options not applying, and the cause would be obvious anyway.

By the way, sorry for the impatience, but with the existing flaws I felt this change could only be an improvement. 😅 And it's working fine on a couple of cameras I had lying around here.

MichaIng: > By the way, sorry for the impatience

No problem at all, sorry for the late review, was busy with another release.

Generally fine to go ahead, when other reviewers take long, while CI and tests are fine. We can always fine tune things in a follow-up, review a final beta/release PR etc. I just didn't want to miss the potential debugging enhancement, now that I found time (and was reminded to) review it.

PR #3407, opened 2026-08-11
fix(v4l2): warn on a failed control listing and don't cache it (follow-up #3391)

Marijn0:
#3391 leaves `vid_control_params` unchanged when `v4l2ctl.list_ctrls()` comes back empty. Since `list_ctrls()` caches its result, losing the connection alone is not enough to trigger this: the cached control IDs are still used when saving. `if video_controls:` can be false in two cases:

- the lookup returned no controls when the page was loaded and still returns none when saving, so no controls are shown in the UI and the existing config values are kept
- the cache was cleared after the page was loaded and the next lookup returns no controls, so the existing `vid_control_params` are kept instead of falling back to the names submitted by the UI

This PR adds a warning when that happens.

This PR also stops `list_ctrls()` from caching an empty result. The cache has no expiry and is only cleared by `list_devices()`, so a failed lookup was kept until that ran or motionEye restarted, leaving the Video Controls section empty long after the device came back. Skipping the cache write means the next lookup simply tries again.

### DIFF
--- a/motioneye/config.py
+++ b/motioneye/config.py
@@ -1131,6 +1131,12 @@ def motion_camera_ui_to_dict(ui, prev_config=None):
                 )
                 data['vid_control_params'] = ','.join(vid_control_params)
 
+            else:
+                logging.warning(
+                    f'camera {prev_config["@id"]} control listing failed, '
+                    'controls not updated'
+                )
+
     else:  # assuming netcam
         if match(
             r'^rtsp|^rtmp', data.get('netcam_url', prev_config.get('netcam_url', ''))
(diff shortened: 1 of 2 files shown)


MichaIng (approved):
That should be fine, also good to not cache empty controls.

Maybe, in a future PR, we can implement some helper function to trigger a non-error/non-fatal frontend/GUI popup from anywhere in backend code. Maybe some sort of global variable that is appended to the JSON response (also 200 ones) by the handlers, and checked by some frontend response handler function, to in case show a popup.

Marijn0:
>Maybe, in a future PR, we can implement some helper function to trigger a non-error/non-fatal frontend/GUI popup

I had exactly the same idea. 😃

PR #3410, opened 2026-08-30
enh(auth): make session lifetime configurable for the normal (surveillance) user

### LINKED ISSUE #3364: 0.44 "remember me" is not shown on login page
After upgrading from 0.43 to 0.44 the "Remember me" is gone.

Image

Tried to find some documentation regarding this , no luck so far.

I assume its something with my settings ?

Any pointers would be great.

Thanks in advance.

Marijn0:
Adds `normal_session_expiry_hours` to motioneye.conf so the session lifetime for the normal (surveillance) user can be configured, with a default of 30 days (720h) and a minimum of 1 hour. As discussed in #3364.

### DIFF
--- a/motioneye/handlers/base.py
+++ b/motioneye/handlers/base.py
@@ -22,7 +22,7 @@ from time import time
 
 from tornado.web import HTTPError, RequestHandler
 
-from motioneye import VERSION, config, template
+from motioneye import VERSION, config, settings, template
 from motioneye.utils.authstate import verify_hmac_signature
 
 __all__ = ('BaseHandler', 'NotFoundHandler', 'ManifestHandler')
@@ -35,12 +35,18 @@ _SESSION_EXPIRY_SECONDS: int = 86400
 _session_store: dict = {}
 
 
+def session_expiry_seconds(user_type):
+    if user_type == 'normal':
+        return max(1, settings.NORMAL_SESSION_EXPIRY_HOURS) * 3600
+    return _SESSION_EXPIRY_SECONDS
+
+
 def create_session(user_type):
     """Create a secure session id with expiry."""
     session_id = token_hex(32)
     _session_store[session_id] = {
         'user': user_type,
-        'expires': time() + _SESSION_EXPIRY_SECONDS,
+        'expires': time() + session_expiry_seconds(user_type),
     }
     return session_id
 
--- a/motioneye/handlers/base.py
+++ b/motioneye/handlers/base.py
@@ -147,7 +147,7 @@ class BaseHandler(RequestHandler):
 
     def get_current_user(self):
         # Check for session-based authentication (via secure cookie)
-        session_id = self.get_secure_cookie('user')
+        session_id = self.get_signed_cookie('user')
         if session_id:
             try:
                 session_id = (
--- a/motioneye/handlers/base.py
+++ b/motioneye/handlers/base.py
@@ -38,7 +38,7 @@ _session_store: dict = {}
 
 def session_expiry_seconds(user_type: str) -> int:
     if user_type == 'normal':
-        return max(1, settings.NORMAL_SESSION_EXPIRY_HOURS) * 3600
+        return max(1, NORMAL_SESSION_EXPIRY_HOURS) * 3600
     return _SESSION_EXPIRY_SECONDS
 
 
--- a/motioneye/handlers/base.py
+++ b/motioneye/handlers/base.py
@@ -22,7 +22,8 @@ from time import time
 
 from tornado.web import HTTPError, RequestHandler
 
-from motioneye import VERSION, config, settings, template
+from motioneye import VERSION, config, template
+from motioneye.settings import NORMAL_SESSION_EXPIRY_HOURS
 from motioneye.utils.authstate import verify_hmac_signature
 
 __all__ = ('BaseHandler', 'NotFoundHandler', 'ManifestHandler')
--- a/motioneye/handlers/base.py
+++ b/motioneye/handlers/base.py
@@ -36,7 +36,7 @@ _SESSION_EXPIRY_SECONDS: int = 86400
 _session_store: dict = {}
 
 
-def session_expiry_seconds(user_type):
+def session_expiry_seconds(user_type: str) -> int:
     if user_type == 'normal':
         return max(1, settings.NORMAL_SESSION_EXPIRY_HOURS) * 3600
     return _SESSION_EXPIRY_SECONDS
--- a/motioneye/handlers/login.py
+++ b/motioneye/handlers/login.py
@@ -24,7 +24,7 @@ from argon2 import PasswordHasher
 from argon2.exceptions import InvalidHash, VerificationError, VerifyMismatchError
 
 from motioneye import config
-from motioneye.handlers.base import BaseHandler, create_session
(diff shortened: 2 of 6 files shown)


motioneye/handlers/base.py
```diff
@@ -35,12 +35,18 @@
...
+    if user_type == 'normal':
+        return max(1, settings.NORMAL_SESSION_EXPIRY_HOURS) * 3600
+    return _SESSION_EXPIRY_SECONDS
+
+
 def create_session(user_type):
     """Create a secure session id with expiry."""
     session_id = token_hex(32)
     _session_store[session_id] = {
         'user': user_type,
```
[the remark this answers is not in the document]
Marijn0: This is actually a pre-existing gap, and I think the in-memory session store we use doesn't align with the practices in Tornado's own documentation: https://www.tornadoweb.org/en/stable/guide/security.html. 
(I need to spend some more time to fully understand why we do things the way we do. 😅)

I can fix the "expired entries only removed when validated" issue if we want, but it's a bit out of scope for this PR, right? And I think that extending the lifetime doesn't make this worse. Besides, only abandoned entries from users who previously authenticated successfully can accumulate, so the practical impact is limited IMO.

zagrim (approved):
👍 
I think the session store bloat is more of a theoretical concern in this case (only two "users", probably pretty low number of actual users creating sessions), and likely only would surface in case of someone actually trying to attack that weakness. Now, with a longer session life-time, there would be more time to grow the store size, but I feel that is still more theoretical than something that is likely happen. Perhaps that issue could be dealt with in another PR later?

motioneye/handlers/base.py
```diff
@@ -22,7 +22,7 @@
 
 from tornado.web import HTTPError, RequestHandler
 
-from motioneye import VERSION, config, template
+from motioneye import VERSION, config, settings, template
```
MichaIng:
```suggestion
from motioneye import VERSION, config, template
from motioneye.settings import NORMAL_SESSION_EXPIRY_HOURS
```

motioneye/handlers/base.py
```diff
@@ -35,12 +35,18 @@
 _session_store: dict = {}
 
 
+def session_expiry_seconds(user_type):
+    if user_type == 'normal':
+        return max(1, settings.NORMAL_SESSION_EXPIRY_HOURS) * 3600
```
MichaIng:
```suggestion
        return max(1, NORMAL_SESSION_EXPIRY_HOURS) * 3600
```

motioneye/handlers/base.py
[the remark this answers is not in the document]
MichaIng: Agreed, we should address this separately, not here.

As far as I know, we could remove our in-memory session store entirely, by just giving the cookie itself a limited lifetime: https://github.com/motioneye-project/motioneye/issues/3364#issuecomment-4763365241
Either the cookie has an inherent lifetime in that case, or tornado handles the database and lifetime validation for us, in a safe way.

motioneye/handlers/login.py
```diff
@@ -143,7 +143,7 @@ def post(self) -> None:
         self.set_secure_cookie(
             'user',
             session_id,
-            expires_days=1,
+            expires_days=session_expiry_seconds(user_type) / 86400,
```
MichaIng, line 146: So, we store the lifetime in the in-memory array in seconds precision, Tornado's `set_secure_cookie` takes the value in days, and the config takes in as hours.

But I see Tornado is accepting a float for `expires_days`, so this should work.

_I just see that it has been renamed to `set_signed_cookie()`, in a version older than what we define as minimum needed: https://github.com/tornadoweb/tornado/commit/49bbd35_

Tornado defaults to 30 days as well, btw: https://github.com/tornadoweb/tornado/blob/master/tornado/web.py#L805

Also interesting for another future PR:

> A value of None limits the lifetime to the current browser session.

So this is a way to re-implement the "remember me" checkbox, which is basically checked hard-coded currently. Leaving it unchecked could then skip the `expires_days` parameter, to let it expire with the browser session.

motioneye/handlers/base.py
```diff
@@ -35,12 +35,18 @@
 _session_store: dict = {}
 
 
+def session_expiry_seconds(user_type):
```
MichaIng:
```suggestion
def session_expiry_seconds(user_type: str) -> int:
```

motioneye/handlers/login.py
```diff
@@ -143,7 +143,7 @@ def post(self) -> None:
         self.set_secure_cookie(
```
MichaIng:
```suggestion
        self.set_signed_cookie(
```
as of https://github.com/tornadoweb/tornado/commit/49bbd35

motioneye/handlers/base.py
MichaIng:
```suggestion
        session_id = self.get_signed_cookie('user')
```
as of https://github.com/tornadoweb/tornado/commit/49bbd35

motioneye/handlers/base.py
Marijn0: I think this change breaks the tests: `from motioneye.settings import NORMAL_SESSION_EXPIRY_HOURS` copies the value once, at import time. It won't pick up later changes from `patch.object()` in tests.

@copilot Can you confirm this is actually the issue, and explain whether this could cause problems in real-world use as well, beyond the failing tests?

motioneye/handlers/login.py
Marijn0: >and the config takes in as hours.

I remember a user request from someone who did not want the session to last even a full day, so I opted to make it configurable in hours.

>So this is a way to re-implement the "remember me" checkbox, which is basically checked hard-coded currently. Leaving it unchecked could then skip the expires_days parameter, to let it expire with the browser session.

It would be nice to bring that option back. I have seen quite a few requests for it, and maybe we could add some help text explaining the current behavior for admin and normal users.

If we bring back the "Remember me" option, configuring the lifetime in hours may no longer be necessary. IMO configuring per hour becomes less useful beyond 24 hours anyway, right? but doesn't hurt either way.😅

tests/test_handlers/test_login.py
```diff
@@ -7,7 +7,8 @@
 import tornado.testing
 from argon2 import PasswordHasher
 
-from motioneye import config, settings
+from motioneye import config
+from motioneye.handlers import base
 from motioneye.handlers.base import _session_store
```
MichaIng, line 12: This looks redundant now (importing `base` as a whole, then `_session_store` from it), but its okay, the tests only. And it will change anyway once we implement Tornado's cookie lifetime handling.

I still wonder whether patching `this` would work as well, since it is a `LoginHandler` instance, which is a `BaseHandler`.

MichaIng (approved):

motioneye/handlers/login.py
MichaIng: Yeah, as it is in hours now, we should leave it like that, else we'd need to keep some backwards compatibility code, or do a migration, and actually the settings file should not even be writable to the service, but R/O. 1 hour is quite a reasonable minimum session lifetime, e.g. for setting up a fresh instance, configuring some cameras, testing things, without the session being lost when closing the browser for some reason. As Tornado does accept a float, that is fine.

MichaIng (approved):

PR #3413, opened 2026-09-09
fix(notifications): send the alert even when the picture listing fails

### LINKED ISSUE #3412: Motion notifications (email + Telegram) silently stop once the picture directory grows: timed-out listing returns `None` and crashes the callback
## Summary

Four related defects. One silently drops notifications, one makes that failure inevitable on any busy camera, one blocks the obvious workaround, and one makes the whole thing invisible in the logs.

**Headline:** on a Raspberry Pi 4 the picture lookup for a single alert takes **19 s against a 10 s timeout**. Bounding that lookup to the time window the caller actually asked for makes it **1 s — a ~19x reduction, with byte-for-byte identical output** (measurements below).

1. **Crash** — `mediafiles.list_media()` resolves its Future with `None` on timeout but with a *list* on success. Both notification senders guard with `if media_files:`, where `media_files` is the **Future object** (always truthy), then iterate `.result()` unguarded. On timeout this raises `TypeError: 'NoneType' object is not iterable` inside an asyncio callback, so **the notification is never sent** and the process still exits 0.
2. **Slowness** — the picture listing is `O(files recorded that day)` when the caller only ever keeps files within `timespan` seconds of the event. With `snapshot_interval` enabled that means building and piping ~17,000 file records to deliver 1–2 of them: **measured at 19 s on a Pi 4, against a 10 s timeout**. Passing the window down into the listing makes the same call take **1 s** and return the same pictures.
3. **Inconsistency** — `sendtelegram.py` never coerces `timespan` to a number, so the string `'0'` is truthy and the "skip picture collection" early-return is unreachable there (it works in `sendmail.py`, which does `int()`).
4. **No visibility** — the senders log all of this correctly, but motion discards their stdout/stderr, so none of it is recorded anywhere. A failing notification leaves no trace at all (details below).

The user-visible effect is notifications that work in the morning and stop partway through the day, recovering by themselves the next morning.

## Symptom

Motion alerts arrive normally early in the day, then stop — reliably around midday on the reporting system — and resume the following morning. Detection is unaffected throughout: events still fire and movies are still recorded. Nothing in the UI indicates a problem; the only trace is a stack trace in the log.

The time-of-day pattern is emergent, not a schedule. The day's picture directory grows continuously (`snapshot_interval 5` produces ~720 files/hour, ~17,000/day), the listing gets slower, and once it exceeds `LIST_MEDIA_TIMEOUT_EMAIL` / `_TELEGRAM` (default **10 s**) every subsequent notification fails. At midnight motion starts a fresh, empty date directory and the cycle repeats.

## Environment

- motionEye **0.44.0** (pip, system Python), motion 4.7.x
- Debian 13 (trixie), Python **3.13.5**
- Raspberry Pi 4 and Pi Zero 2 W, microSD storage
- `snapshot_interval 5`, `picture_filename %Y-%m-%d/%H-%M-%S`, email + Telegram notifications enabled

### Defect 1 — the guard tests the Future, not the result

`mediafiles.list_media()`, in `poll_process`:

```python
else:  # process did not finish in time
    logging.error('timeout waiting for the media listing process to finish')
    try:
        os.kill(process.pid, SIGKILL)
    except:
        pass  # nevermind

    fut.set_result(None)          # <-- timeout

else:  # finished
    read_media_list()
    fut.set_result(media_list)    # <-- success
```

`sendmail.py` (`make_message`, ~line 84) and `sendtelegram.py` (~line 66) both then do:

```python
def on_media_files(media_files):
    io_loop.stop()
    timestamp = time.mktime(moment.timetuple())

    if media_files:                        # Future object -> always True
        logging.debug('got media files')
        media_files = [
            m
            for m in media_files.result()  # None on timeout -> TypeError
        ]
```

That the guard misfires is directly observable: `got media files` is logged **immediately after** the timeout error, which can only happen if `if media_files:` passed on a timed-out listing.

```
[sendmail]    DEBUG: starting media listing subprocess...
[sendmail]    ERROR: timeout waiting for the media listing process to finish
[sendmail]    DEBUG: got media files          <-- guard wrongly passed
[sendmail]    ERROR: Exception in callback make_message.<locals>.on_media_files()
  File "/usr/lib/python3.13/asyncio/events.py", line 89, in _run
    self._context.run(self._callback, *self._args)
  File ".../motioneye/sendmail.py", line 95, in on_media_files
    for m in media_files.result()
             ~~~~~~~~~~~~~~~~~~^^
TypeError: 'NoneType' object is not iterable
```

Identical trace from `sendtelegram.py:74`.

Note `on_media_files` is called two ways — with a Future from `add_done_callback`, and with a plain list from the `if not timespan: return on_media_files([])` early return — so any fix must accept both.

### Defect 2 — the listing is unbounded

The senders keep only files satisfying `abs(m['timestamp'] - event_ts) < timespan`, yet `_do_list_media` formats and ships **every** file first. Per file it builds a dict containing `guess_type()`, **two** `pretty_date_time()` calls, `pretty_size()` and two `datetime.fromtimestamp()` constructions, then `pipe.send()`s it individually — while the parent drains that pipe only from an IOLoop timer firing every **0.5 s** (`read_media_list` inside `poll_process`).

On a 17,280-file day directory that is roughly 630 records/s, i.e. **~27 s of IPC alone**, before disk cost. The 10 s default is exceeded outright.

## Measurements

All figures below are from a **Raspberry Pi 4 Model B** (4 cores, 3.8 GB RAM) — not
constrained hardware — against a **real 17,450-file day directory**, page cache dropped
(`sync; echo 3 > /proc/sys/vm/drop_caches`) so the run resembles an event after an idle period.

**Reproduction of the failure at the shipped default:**

```
# with list_media_timeout_email at its default of 10
DEBUG: starting media listing subprocess...
ERROR: timeout waiting for the media listing process to finish
TypeError: 'NoneType' object is not iterable          <-- notification lost
```

**The same listing, allowed to finish:**

```
DEBUG: narrowing down still images path lookup to 2026-09-07
DEBUG: starting media listing subprocess...            02:11:37
DEBUG: media listing process has returned 17266 files  02:11:57   <-- 20 seconds
DEBUG: selected 4 pictures
```

So on a Pi 4 the listing takes **20 s against a 10 s budget** — 2x over — and delivers
**4 useful pictures out of 17,266**.

**Where that 20 s goes.** A plain `os.scandir()` + `entry.stat()` walk over the same
directory on the same machine, also cold, takes **0.84 s**:

| Work | Pi 4 | Pi Zero 2 W |
|---|---|---|
| `scandir` + `stat` over the whole directory | 0.84 s | 1.45 s |
| motionEye's full `list_media()` | **20 s** | — |

The stat walk is therefore only ~4 % of the cost. The other ~96 % is the per-file dict
construction (`guess_type()`, two `pretty_date_time()` calls, `pretty_size()`, two
`datetime.fromtimestamp()`), the pickling, and ~17,000 individual `pipe.send()` calls drained
by the parent from an IOLoop timer that fires every 0.5 s.

**Measured with the fix applied**, same machine, same directory, same event, cold cache:

```
# stock
DEBUG: starting media listing subprocess...            08:59:35
DEBUG: media listing process has returned 17266 files  08:59:54   <-- 19 s
DEBUG: selected 4 pictures

# with the time window pushed into the listing
DEBUG: starting media listing subprocess...            09:00:10
DEBUG: media listing process has returned 4 files      09:00:11   <-- 1 s
DEBUG: selected 4 pictures
```

Identical output — the same four pictures are selected — in **1 s instead of 19 s, a ~19x reduction**, because
17,266 records are no longer built and piped only to be discarded. That puts the existing 10 s
default back in comfortable range with no configuration change.

This is not a slow-hardware problem: it reproduces on a Pi 4 and would reproduce on any
machine once the directory is large enough, because the cost scales with files-recorded-that-day
rather than with the handful of files the caller actually wants.

### Defect 4 — none of the above is observable

The senders already log what an operator would need: `sending email`, `email sent`,
`timeout waiting for the media listing process to finish`, and the traceback. But motion
discards the children's stdout/stderr, so **none of it reaches the journal, the motionEye log,
or anywhere else**. On the reporting system, seven days of logs contained zero occurrences of
`email sent`, `sending email`, `timeout waiting`, or `NoneType`.

The practical effect is that a multi-week notification outage was undiagnosable after the fact.
The failure had to be inferred from user-visible behaviour (alerts arriving in the morning and
not the afternoon) and then reproduced by running the sender by hand. Note also that
`meyectl` reports `configure logging to file: None` for these subcommands, so `log_path` in
`motioneye.conf` does not capture them either.

This is a small thing to fix and would have turned a multi-day investigation into a single
`grep`: having the senders record a one-line outcome per attempt (sent / no pictures /
listing timed out / send failed) somewhere durable would be enough.

One caveat worth flagging for whoever implements it: the senders receive the SMTP password and
the Telegram bot token as command-line arguments. At the default log level neither is echoed,
but with `-d` `sendtelegram` logs an argparse `Namespace` containing the API key in full, so any
new logging should be careful not to capture argv or raw output.

## Reproduction

1. Configure a camera with `snapshot_interval 5` and `picture_filename %Y-%m-%d/...`.
2. Enable motion notifications (email and/or Telegram) with a non-zero picture time span (the UI default emits `... '%Y-%m-%dT%H:%M:%S' '10'`).
3. Let the current day's picture directory reach roughly 10 k+ files.
4. Trigger motion, or invoke directly:

```
meyectl sendmail -d -c /etc/motioneye/motioneye.conf \
    <server> <port> <account> <password> <tls> <from> <to> \
    'motion_start' <camera_id> "$(date '+%Y-%m-%dT%H:%M:%S')" '10'
```

Expected: a notification is sent, with or without pictures.
Actual: `TypeError: 'NoneType' object is not iterable`; no notification.

The same command succeeds earlier in the day against a smaller directory, which is what makes the failure look time-based.

## Proposed fix

Three small, independent changes. **The first is the important one** — it converts a total notification failure into a picture-less notification, and is worth landing on its own regardless of the rest.

**1. Don't lose the notification when the listing fails.** Resolve the Future (handling the dual list/Future calling convention), treat `None` as "no pictures", log a warning, and send anyway. Duck-typing on `.result` avoids adding a `Future` import to either module.

**2. Coerce `timespan` in `sendtelegram.py`** (guarded), mirroring `sendmail.py`, so `timespan=0` disables picture collection consistently.

**3. Push the caller's time window into the listing.** Add keyword-only `min_timestamp` / `max_timestamp` to `list_media()`, thread them through `_do_list_media()` into `_list_media_files()`, and skip out-of-window files immediately after `entry.stat()` — before the dict, the formatting, the pickle and the `pipe.send()`. Bounds are inclusive while the senders keep their existing strict `<` filter unchanged, so the pre-filter is strictly wider than the caller's own test and **the delivered file set is provably identical**.

A complete BEFORE/AFTER patch for all three is attached/linked.

Two details a reviewer should note in change 3:

- `_list_media_files` **recurses into subdirectories** passing only `with_stat` by keyword. A new parameter that isn't explicitly forwarded there is silently ignored for every subdirectory (i.e. whenever `prefix` is None), which would look like a working filter that quietly returns unfiltered results from nested directories.
- With no window passed, `windowed` is False and the appended tuple is unchanged, so the no-window path is identity by inspection. The other `_list_media_files` consumers — `_remove_older_files` (the cleanup janitor), `_do_zip`, `_do_list_pictures`, `del_media_group` — are unaffected.

## Not addressed here

- The 0.5 s `poll_process` cadence and per-file `pipe.send()` remain, so the **web media browser** still pays the full cost on large directories. Batching sends into sub-`PIPE_BUF` chunks would help that caller; it is a separate concern with a different risk profile, and is deliberately left out.
- `handlers.py` and `remote.py` were not available to the author. Please grep for `list_media(` before merging change 3 to confirm no caller passes a 5th positional argument (the `*` would make that a `TypeError`); a 4-positional call such as `list_media(cfg, 'picture', prefix, False)` still works. Remote/proxied listings do not forward the new parameters and keep today's cost.
- Pre-existing and untouched: motion writes JPEGs directly to their final name with no temp+rename, so a picture still being written can be attached truncated; and `time.mktime(moment.timetuple())` uses `tm_isdst=-1`, which can resolve the repeated local hour of a DST fall-back to the wrong side by an hour.

## Interim workaround

Until this is fixed, raising the notification listing timeouts in `motioneye.conf` restores alerts:

```
list_media_timeout_email 120
list_media_timeout_telegram 120
```

This is a mitigation only — the directory keeps growing, so any fixed value can eventually be exceeded, and alert latency grows through the day. (Reducing `snapshot_interval` has the same mitigating effect by keeping the directory small.)

---

*The diagnosis and all measurements above were performed on a live 0.44.0 installation. The proposed code has not been run against the upstream test suite.*

free-cup-holder-exe:
Fixes the crash half of #3412.

`mediafiles.list_media()` resolves its Future with `None` on timeout but with a list on
success. Both notification senders guard with `if media_files:` — where `media_files` is the
**Future object**, which is always truthy — and then iterate `.result()` unguarded. On a
timeout that raises `TypeError: 'NoneType' object is not iterable` inside an asyncio callback,
so the notification is silently never sent while the process still exits 0.

That the guard misfires is directly observable: `got media files` is logged immediately after
the timeout error, which can only happen if the guard passed on a timed-out listing.

`on_media_files` is called both with a Future (from `add_done_callback`) and with a plain list
(the `if not timespan: return on_media_files([])` early return), so this resolves it
accordingly, treats `None` as "no pictures", logs a warning naming the relevant timeout
setting, and sends the notification without attachments rather than dropping it. Duck-typing on
`.result` avoids adding a `Future` import to either module.

Also coerces `sendtelegram`'s `timespan` to `int`, as `sendmail` already does — without it the
CLI string `'0'` is truthy, so the `if not timespan:` early return meant to disable picture
collection is unreachable there.

This is deliberately scoped to the crash only. The underlying reason the listing times out (it
is `O(files recorded that day)` for a caller that keeps 1–2 files) is described in #3412 and
addressed separately, so this one can be reviewed on its own.

### Verification

- **Full test suite passes**: 136 tests, and `tests/test_mediafiles.py` on its own (38 tests).
  Run on a Raspberry Pi 4 (the suite needs a Unix host — `mediafiles` imports `fcntl`).
- **Lint clean** using this repo's own `.pre-commit-config.yaml` arguments — `black`
  (`--skip-string-normalization`, all `-t` targets), `flake8` (`--max-line-length=125`,
  `--select=C901,E501,E9,F63,F7,F82`) and `isort` (`--profile black`) all report no change
  against these files, and introduce no new findings relative to `dev`.
- **Behaviour**: reproduced the original failure on motionEye 0.44.0 / Debian 13 /
  Python 3.13.5 by letting a camera's day picture directory reach ~17k files, then confirmed
  the notification is delivered (without attachments) instead of being dropped.

No documentation or changelog changes appear to be required (releases are used for the
changelog, and there is no PR template or sign-off requirement in the repo).


motioneye/sendmail.py
```diff
@@ -94,14 +94,21 @@ def on_media_files(media_files):
 
         timestamp = time.mktime(moment.timetuple())
 
-        if media_files:
+        files = media_files.result() if hasattr(media_files, 'result') else media_files
```
MichaIng: Let's start to use typing when declaring new variables:
```suggestion
        files: list = media_files.result() if hasattr(media_files, 'result') else media_files
```

motioneye/sendtelegram.py
```diff
@@ -67,12 +67,19 @@ def on_media_files(media_files):
         io_loop.stop()
 
         timestamp = time.mktime(moment.timetuple())
-        if media_files:
+        files = media_files.result() if hasattr(media_files, 'result') else media_files
```
MichaIng:
```suggestion
        files: list = media_files.result() if hasattr(media_files, 'result') else media_files
```

motioneye/sendmail.py
```diff
@@ -94,14 +94,21 @@ def on_media_files(media_files):
...
+        if files is None:
+            logging.warning(
+                'picture listing timed out after %ss; sending notification '
+                'without pictures (consider raising list_media_timeout_email)',
+                settings.LIST_MEDIA_TIMEOUT,
```
MichaIng: IMO f-strings are easier to read:
```suggestion
                f'picture listing timed out after {settings.LIST_MEDIA_TIMEOUT}s; sending '
                'notification without pictures; consider raising list_media_timeout_email',
```

motioneye/sendtelegram.py
```diff
@@ -67,12 +67,19 @@ def on_media_files(media_files):
...
+        if files is None:
+            logging.warning(
+                'picture listing timed out after %ss; sending notification '
+                'without pictures (consider raising list_media_timeout_telegram)',
+                settings.LIST_MEDIA_TIMEOUT,
```
MichaIng:
```suggestion
                f'picture listing timed out after {settings.LIST_MEDIA_TIMEOUT}s; sending '
                'notification without pictures; consider raising list_media_timeout_email',
```

motioneye/sendtelegram.py
```diff
@@ -67,12 +67,19 @@ def on_media_files(media_files):
...
+            )
+            files = []
+
+        if files:
             logging.debug('got media files')
             media_files = [
-                m
-                for m in media_files.result()
-                if abs(m['timestamp'] - timestamp) < float(timespan)
+                m for m in files if abs(m['timestamp'] - timestamp) < float(timespan)
```
MichaIng: Do I see it right that `timespan` is assured to be an integer now, so that we do not need to cast it anymore? There are more cases below.
```suggestion
                m for m in files if abs(m['timestamp'] - timestamp) < timespan
```
Might go beyond the scope of this PR, but generally makes sense to make this clear with function arg typing `timespan: int` already.

MichaIng (commented):
Many thanks, both changes make sense. Some minor remarks, with the option to apply some more typing across the functions/script, to prevent exactly such kind of ambiguous values and related errors.

mypy by default skips type checking if the functions themselves are not typed (args and/or return value), but it does not hurt to start using it consequently for any newly added variable and function definitions.

motioneye/sendmail.py
free-cup-holder-exe: Agreed on the typing — applied, with one small deviation worth flagging.

Annotating the raw assignment as `list` would be inaccurate here: that expression is precisely the one that can yield `None` (the timeout case this PR fixes), so `files: list = ...` would assert something the very next line disproves.

So I kept the annotation but moved it past the guard:

```python
raw_files = (
    media_files.result() if hasattr(media_files, 'result') else media_files
)
if raw_files is None:
    logging.warning(...)
files: list = raw_files or []
```

`files` is genuinely a `list` at the point it's annotated, and `or []` folds the `None` and empty-list cases together — which the following `if files:` already treated identically.

Happy to switch to `Optional[list]` on the original line instead if you'd rather the annotation sit exactly where you put it.

motioneye/sendtelegram.py
free-cup-holder-exe: Applied the same way as in `sendmail.py` — see the note on that thread. The annotation moved one line down, past the `None` guard, so `files: list` is accurate rather than aspirational.

motioneye/sendmail.py
free-cup-holder-exe: Applied — agreed, they read better.

One change beyond the suggestion: the f-string makes the trailing `settings.LIST_MEDIA_TIMEOUT,` positional argument redundant, so I removed it as well. Taking the suggestion on its own would have left that line orphaned as an unused `logging` argument.

motioneye/sendtelegram.py
free-cup-holder-exe: Applied, with one correction: this is the Telegram sender, so the setting to point at is `list_media_timeout_telegram`, not `list_media_timeout_email` — I think the `_email` name carried over from the `sendmail.py` suggestion.

It matters here because `main()` in this file overrides `settings.LIST_MEDIA_TIMEOUT` with `LIST_MEDIA_TIMEOUT_TELEGRAM`, so naming the email knob would send people to a setting that has no effect on the timeout they just hit.

Same removal of the now-redundant `settings.LIST_MEDIA_TIMEOUT,` argument as in `sendmail.py`.

motioneye/sendtelegram.py
free-cup-holder-exe: Yes, that's right. This PR adds to `sendtelegram.main()` the same `int()` coercion `sendmail.py` already did:

```python
try:
    options.timespan = int(options.timespan)
except (TypeError, ValueError):
    options.timespan = 0
```

so `timespan` is an `int` by the time `make_message()` sees it, and all three casts are redundant. I removed all of them — this comprehension, plus the `logging.debug` and `time.sleep` further down.

Worth noting the coercion fixes a second bug on its own. `argparse` hands the value over as a string, so `timespan` was `'0'` — truthy — which made the existing early return

```python
if not timespan:
    return on_media_files([])
```

unreachable from the CLI. Setting the picture time span to 0 disabled picture collection for email but not for Telegram, which still ran the full (potentially slow) listing.

On the typing suggestion: added `timespan: int` to both `make_message()` signatures, along with the remaining parameters.

One thing I deliberately left alone, in case you would rather handle it separately: I typed `camera_id: int`, but `motionctl.motion_camera_id_to_camera_id()` returns `None` on `IndexError`, so strictly it is `Optional[int]` at the call site. `config.get_camera(None)` would fail regardless, so this looks like a latent bug rather than a supported input — widening the annotation to `Optional[int]` felt like it would document the bug rather than fix it.

motioneye/sendmail.py
MichaIng: Ah right, `Optional[list]` was also my first thought when seeing the code change.

Or, to merge things, to avoid an intermediate variable, and have a strict type right from the start:
```py
files: list = media_files.result() or [] if hasattr(media_files, 'result') else media_files
```

motioneye/sendmail.py
MichaIng: Right, the trailing `,` was more an oversight, being used to keep the trailing comma in lists in other programming languages, even if no further element is coming (same like there was one after `settings.LIST_MEDIA_TIMEOUT,`). But here it is cleaner to not indicate/imply any use of further positional arguments.

motioneye/sendtelegram.py
MichaIng: We will see whether mypy throws an error as of the `camera_id` type mismatch. If so, we can think about whether to allow `Optional[int]` here or throw a meaningful error at the caller side.

motioneye/sendmail.py
MichaIng: I just recognize that this would break the ability to throw a meaningful warning if the listing timed out. So assigning and handling None at first makes sense. `Optional[list]` or separate raw and list variables does not really make a difference in terms of amount of code or complexity. The only reason for `Optional[list]` would be to cause a mypy error if `media_files.result()` or `media_files` have any other falsely type than None, once those are further typed throughout the code.

free-cup-holder-exe: Agreed on both counts — and yes, the merged form loses the warning, which is the only reason I split it.

On `Optional[list]` versus the separate `raw`: I'll happily switch if you want the stricter signal. The reason I landed on `raw` in the end is that `files: list = media_files.result() if hasattr(media_files, 'result') else media_files` is 93 characters at that indentation, so black wraps it into a parenthesised block either way — whereas `raw = ...` fits at 85 and stays on one line. Purely a formatting side effect, not a semantic preference.

Your point about the future mypy value is the stronger argument though: with `Optional[list]` a non-`None` falsy result would be caught once `list_media()` and friends are typed, while `raw or []` would quietly swallow it. Say the word and I'll make the swap.

motioneye/sendtelegram.py
free-cup-holder-exe: Worth noting the current head no longer annotates `camera_id` — when I collapsed the signatures back to one line I dropped everything except the `timespan: int` you asked for, so there is no type mismatch for mypy to find at the moment.

The underlying issue is still real for a later typing pass: `motionctl.motion_camera_id_to_camera_id()` returns `None` on `IndexError`, so a strict signature would be `Optional[int]`. For what it is worth I think the meaningful-error-at-the-caller option is the better of the two you listed — `config.get_camera(None)` fails anyway, just further away from the cause, so `Optional[int]` would mostly document the bug rather than fix it.

motioneye/sendmail.py
```diff
@@ -94,14 +94,20 @@ def on_media_files(media_files):
...
+
+        if files:
             logging.debug('got media files')
 
             # filter out non-recent media files
             media_files = [
-                m
-                for m in media_files.result()
-                if abs(m['timestamp'] - timestamp) < timespan
+                m for m in files if abs(m['timestamp'] - timestamp) < timespan
```
Demetriao: Bug: empty media results leave the original `Future` in `media_files`, which is then passed to the callback

Cause:

`raw`/`files` correctly unwrap the `Future` and normalize a timeout (`None`) to an empty list. However, `media_files` is only reassigned inside `if files:`:

```python
if files:
    media_files = [m for m in files if ...]
    media_files.sort(...)
    media_files = [...]
# No else branch.
```

When the listing returns `None` or `[]`, this block is skipped, leaving `media_files` as the original `Future`.

Effect:

The callback receives that Future instead of an empty list. The downstream `files or []` does not help because the Future is truthy.

- **sendmail.py:** `reversed(files)` raises `TypeError`. The email is not sent, and a send failure is logged.
- **sendtelegram.py:** `for f in files:` iterates the completed `asyncio.Future` and yields no items. No message or photo is sent, but `"telegram sent"` is still logged.

The timeout fallback therefore remains broken. An ordinary empty listing also becomes a regression: it previously allowed a notification without pictures.

Suggestion:

Add the missing `else` branch in both files so the callback receives an empty list when no pictures are available:

```python
if files:
    media_files = [m for m in files if ...]
    media_files.sort(...)
    media_files = [...]
else:
    media_files = []
```

motioneye/sendmail.py
MichaIng: Let's prefer stricter typing over purely visual formatting concerns, and go with `Optional[list]`.

motioneye/sendmail.py
MichaIng: It would be probably cleaner to not use `media_files` anymore after `files: list` has been assigned, but `files`, which is assured to be a list.

free-cup-holder-exe: Confirmed, and thank you — this is exactly right, on both counts. I reproduced it against the real code rather than only reading it, driving `make_message()` with a stubbed listing and recording what actually reaches the callback:

| listing result | 0.44.0 | this PR, before your report |
| --- | --- | --- |
| `None` (timed out) | `TypeError` in the comprehension, notification lost | callback receives the **Future**, `reversed()` raises, notification lost |
| `[]` (no pictures) | callback receives `[]`, **notification sent** | callback receives the **Future**, `reversed()` raises, **notification lost** |
| populated | sent | sent |

So the first column of the timeout row was the bug I set out to fix, and I had only moved it from the comprehension into `reversed()` — the notification was still lost. And you are right that the empty-listing row was a straight regression against 0.44.0, which is worse, because that path previously worked.

Fixed. Rather than adding an `else` branch I took @MichaIng's suggestion above and stopped using `media_files` after the unwrap entirely, so there is no second name that can go stale:

```py
files: Optional[list] = (
    media_files.result() if hasattr(media_files, 'result') else media_files
)
if files is None:
    logging.warning(...)
    files = []

if files:
    files = [m for m in files if abs(m['timestamp'] - timestamp) < timespan]
    files.sort(key=lambda m: m['timestamp'], reverse=True)
    files = [os.path.join(...) for m in files]
...
callback(s, m, files)
```

`media_files` now appears only as the parameter and in the unwrap expression. Same change in `sendtelegram.py`.

Re-running the same harness on the corrected code, all three rows now hand the callback a `list` and the notification is sent. The full suite still passes.

motioneye/sendmail.py
free-cup-holder-exe: Done — switched to `Optional[list]` on the unwrap, with the `None` normalised to `[]` immediately after the warning. Agreed that the stricter signal is worth more than keeping the line unwrapped.

Worth flagging that this thread's discussion is what surfaced the real bug: while making the annotation accurate I had left `media_files` in use further down, which is the defect @Demetriao reported below.

MichaIng (approved):
LGTM, no further suggestions.

The `mypy` warnings, that the body of functions which have no arg or return value typing, are not type checked either, are expected. We could enforce it, but that will most likely result in a large amount of annotations. Easiest to apply consequent typing for those functions we touch in other contexts, and address mypy annotations then, when we need to understand and test the functions and their callers anyway. A gradual long-term task 🙂.

